PR #14, opened 2019-05-21
Corrected issues with services, service accounts and sccs

sabre1041:
Updates to the following:

* Corrected PostgreSQL service port number
* Added new service account for redis and realigned quayecosystem->quay
* Adds multiple service accounts to the `anyuid` scc


thomasmckay:
LGTM

PR #20, opened 2019-06-10
Separated out imagePullSecrets and added support for registry storage

sabre1041:
* Allow for specifying an ImagePullSecret per resource
* Support for managing registry storage (currently supports local type)


thomasmckay (approved):
LGTM
Thanks for updating the README as well.

PR #25, opened 2019-07-18
Registry and certificate management enhancements

sabre1041:
* Restructured how to handle persistent storage for the backends
    * Will enable adding other provider much easier
* Initial certificate managemnt
    * Creation of public/private SSL keys


thomasmckay:
LGTM - visual ack

Issue #41, opened 2019-08-09
Add support for deploying Clair


sabre1041:
Integration with Clair as part of the operator

sabre1041:
Closed via #42

Issue #40, opened 2019-08-06
using credentialsSecretName breaks deployment


itewk:
when I specify a custom `credentialsSecretName` the operator fails to deploy the DB.

operator logs:

```
"level":"error","ts":1565109241.4033964,"logger":"controller_quayecosystem","msg":"Failed to Setup Quay","error":"database Validation Failed: FATAL:  password authentication failed for user \"quay\"\n","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/travis/gopath/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128\ngithub.com/redhat-cop/quay-operator/pkg/controller/quayecosystem.(*ReconcileQuayEcosystem).Reconcile\n\t/home/travis/gopath/src/github.com/redhat-cop/quay-operator/pkg/controller/quayecosystem/quayecosystem_controller.go:191\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/travis/gopath/pkg/mod/sigs.k8s.io/controller-runtime@v0.1.10/pkg/internal/controller/controller.go:215\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/home/travis/gopath/pkg/mod/sigs.k8s.io/controller-runtime@v0.1.10/pkg/internal/controller/controller.go:158\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/travis/gopath/pkg/mod/k8s.io/apimachinery@v0.0.0-20181127025237-2b1284ed4c93/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/travis/gopath/pkg/mod/k8s.io/apimachinery@v0.0.0-20181127025237-2b1284ed4c93/pkg/util/wait/wait.go:134\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/travis/gopath/pkg/mod/k8s.io/apimachinery@v0.0.0-20181127025237-2b1284ed4c93/pkg/util/wait/wait.go:88"}
{"level":"info","ts":1565109281.409356,"logger":"controller_quayecosystem","msg":"Reconciling QuayEcosystem","Request.Namespace":"quay-enterprise","Request.Name":"quay"}
```

secrete creation:
```
[...]

PR #42, opened 2019-08-11
Clair integration

### LINKED ISSUE #32: Need a way to customize the route
Right now the default route is always used for Quay. Which makes for a rather "funny" URL. For a quay-enterprise installation of name 'quay' it is quay-quay-quay-enterprise.apps.cluster-shared.shared.example.opentlc.com.

It would be good to just specify the route name as "quay.apps.cluster-shared.shared.example.opentlc.com".

So instead of service-project (which is quay-quay - quay-enterprise) just 'quay' would be enough.

Add a field to the CR to change this.

sabre1041:
Clair integration

Resolves #32 #41 #40


wkulhanek:
Seems to work fine. Except Clair lets every image I throw at it pass as 100%.

sabre1041:
@wkulhanek added an `updateInterval` flag to reduce the 100% pass rate

thomasmckay (approved):
LGTM

PR #70, opened 2019-09-13
Providing basic examples 

scouturier:
Providing basic examples based on real-life usage


cnuland (approved):
LGTM

sabre1041 (dismissed):
LGTM

deploy/examples/basic_automated_cr.yml
```diff
@@ -0,0 +1,18 @@
...
+  name: quayecosystem
+spec:
+  quay:
+    imagePullSecretName: redhat-pull-secret
+    deploymentStrategy: Recreate
+    skipSetup: false
+    keepConfigDeployment: true
+    routeHost: quay.app.example.com
+  redis:
+    imagePullSecretName: redhat-pull-secret
```
sabre1041: Typically, Redis would originate from a different registry than Quay/Clair (Red Hat Container Catalog vs Quay). By default, the operator specified _registry.access.redhat.com_ as the registry hosting Redis and this registry does not require authentication

sabre1041 (changes requested):

deploy/examples/basic_automated_cr.yml
sabre1041: @scouturier bump ^^

scouturier: I have updated both files to use the red hat repo rather than quay.

sabre1041 (approved):
LGTM

PR #85, opened 2019-10-25
Added support for S3 storage

sabre1041:
Support for S3 registry storage. Follow steps listed in the `README.md` to make use of storage


pkg/controller/quayecosystem/constants/constants.go
```diff
@@ -148,6 +148,15 @@ const (
 	// RegistryStorageTypeLocalStorageName is the value of the Local Quay Storage type
 	RegistryStorageTypeLocalStorageName = "LocalStorage"
 
+	// RegistryStorageTypeS3StorageName is the value of the Amazon S3 storage type
+	RegistryStorageTypeS3StorageName = "S3Storage"
+
+	// S3AccessKey is the name of the S3 Access Key
+	S3AccessKey = "s3_access_key"
+
+	// S3SecretKey is the name of the S3 Access Key
```
kurtismullins, line 157: s/Access/Secret

pkg/controller/quayecosystem/constants/constants.go
```diff
@@ -230,6 +239,9 @@ var (
 	// RequiredSslCertificateKeys represents the keys that are required for a provided SSL certificate
 	RequiredSslCertificateKeys = []string{QuayAppConfigSSLCertificateSecretKey, QuayAppConfigSSLPrivateKeySecretKey}
 
+	// RequiredS3CredentialKeys rerpresents the keys that are required for the S3 registry backend
```
kurtismullins, line 242: s/rerpresents/represents

kurtismullins (approved):
Overall, the S3 configuration appears to be functional and the documentation is both clear and accurate.

I noticed that the operator will reconcile the S3 credentials when I moved them from my CustomResource to a secret and then `oc apply` my updated CustomResource. However, the same behavior did not occur when I attempted to add an S3 replica to my CustomResource. I do not consider this a blocker.

When instantiating a new Quay environment with storage replication configured, I was not able to verify that the storage replication is working correctly. It appears the operator generates a valid Quay `config.yaml` and I see no obvious errors in my pod's logs. The second S3 bucket simply contains a file called `_verify` and nothing more appears to be copied over by the storagereplication worker. I assume this is an issue related to my environment and not the operator.

PR #119, opened 2019-12-22
Feature to enable repository mirroring

### LINKED ISSUE #64: As a user, I want to run repo mirroring services, so that the feature is enabled and works.
With quay-v3.1.0 there is a new feature named "repo mirroring". This feature defaults to off and thus must be enabled via the config.yaml/config app. In addition, by default the worker (aka service) that does the mirroring is not started in the "registry" container mode. Instead, a new entrypoint "repomirror" is available to run only the services necessary for this feature. The intent is that there will be both a "registry" pod and the "repomirror" pod running. The "repomirror" pod requires access to the database (like "registry" does), as well as access to external registries to be mirrored. An external endpoint is not required.

An alternate starting mechanism for "repomirror" is available to run the worker inside the "registry" container. To do this, the environment variable ```QUAY_OVERRIDE_SERVICES``` is given a value of ```repomirrorrowker=true```. This mode is intended to be for POC but is fully functional and may, in fact, be the most common mode for users.

For the quay-operator, I'd suggest two phase approach: First, add the env var method to allow usage. Next, add effectively a duplicate setup as "registry" but change entry point to "repomirror".

Differences in "repomirror" pod:

- At this time, there is no readiness/liveness endpoints so default them to be omitted. Leave the functionality, though, since upcoming release will support that.
- An external route to this pod is unnecessary at this time. A distant release may support this, though, so if it makes sense to leave support in code but disabled, that would not be wasted effort.
- Scaling, sizing, and other pod aspects should be available but separate from the "registry" pods (ie. they will be adjusted separately).

sabre1041:
Feature to enable repository mirroring to resolve #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,27 @@ Quay supports multiple storage backends (configured as an array). The quay opera
 
 Please refer to the [Registry Storage Documentation](docs/storage.md) for the options available.
 
+### Repository Mirroring
+
+Quay provides the capability to create container image repositories that exactly match the content of external registries. This functionality can be enabled by setting the `enableRepoMirroring: true` as shown below:
+
+```
+apiVersion: redhatcop.redhat.io/v1alpha1
+kind: QuayEcosystem
+metadata:
+  name: example-quayecosystem
+spec:
+  quay:
+    imagePullSecretName: redhat-pull-secret
+    enableRepoMirroring: true
+```
+
+The following additional options are also available:
+
+* `repoMirrorTLSVerify` - Require HTTPS and verify certificates of Quay registry during mirror
+* `repoMirrorServerHostname` - URL for use by the skopeo copy command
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -212,6 +212,27 @@ Quay supports multiple storage backends (configured as an array). The quay opera
...
+  name: example-quayecosystem
+spec:
+  quay:
+    imagePullSecretName: redhat-pull-secret
+    enableRepoMirroring: true
+```
+
+The following additional options are also available:
+
+* `repoMirrorTLSVerify` - Require HTTPS and verify certificates of Quay registry during mirror
```
kurtismullins, line 232: @sabre1041 
It may help to word this in a way that explains it toggles verification of the external registry's certificate as well

kurtismullins: Scratch that. This is accurate.

kurtismullins (approved):
LGTM

PR #120, opened 2019-12-23
Corrected credential storage keys

sabre1041:
Corrected the keys that are found in secrets representing storage credentials

For GA (1.0.0), to make use of credentials stored in secrets, the values that are being replaced in this PR will need to be used. This will need to be documented. @kurtismullins @thomasmckay any thoughts on how we would want to do this (Storage document?)


kurtismullins:
@sabre1041 

>any thoughts on how we would want to do this (Storage document?)

I think the storage document works great for now. I like the format you've started with; demonstrating how to configure each storage backend individually.

kurtismullins (approved):
Only verified S3 but in general it LGTM.

PR #128, opened 2020-01-06
E2E Tests

cnuland:


pkg/controller/quayecosystem/provisioning/provision.go
```diff
@@ -927,7 +927,7 @@ func (r *ReconcileQuayEcosystemConfiguration) ManageQuayEcosystemCertificates(me
 
 			clairSslSecret = resources.GetTLSSecretDefinition(meta, privKeyBytes, certBytes)
 
-			err = r.reconcilerBase.CreateOrUpdateResource(r.quayConfiguration.QuayEcosystem, r.quayConfiguration.QuayEcosystem.Namespace, clairSslSecret)
+			err = r.reconcilerBase.CreateResourceIfNotExists(r.quayConfiguration.QuayEcosystem, r.quayConfiguration.QuayEcosystem.Namespace, clairSslSecret)
```
cnuland: Found a bug where if the operator triggered a reconcile twice it would create two instances of the same pod. I believe the desired functionality is to have 1 pod, @sabre1041 please review.

pkg/controller/quayecosystem/resources/deployments.go
```diff
@@ -93,6 +93,10 @@ func GetQuayConfigDeploymentDefinition(meta metav1.ObjectMeta, quayConfiguration
 	BuildQuayConfigResourceLabels(meta.Labels)
 
 	envVars := []corev1.EnvVar{
```
cnuland, line 95: This is required in the latest version of Quay, 3.2.2

pkg/k8sutils/k8sutils.go
```diff
@@ -29,55 +31,34 @@ func GetK8sClient(config *rest.Config) (kubernetes.Interface, error) {
...
-		logging.Log.Error(err, "An error occurred")
-	}
-	ch := watcher.ResultChan()
-	logging.Log.Info("Waiting for deployment", "Namespace", namespace, "Name", name, "Timeout", timeout)
-	for event := range ch {
-		dc, ok := event.Object.(*appsv1.Deployment)
-		if !ok {
-			logging.Log.Error(err, "Unexpected type")
-		}
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
```
cnuland: Ripped out the need for the K8Client and instead am using the operator sdk client

pkg/controller/quayecosystem/quayecosystem_controller.go
```diff
@@ -177,11 +183,12 @@ func (r *ReconcileQuayEcosystem) Reconcile(request reconcile.Request) (reconcile
 			return r.manageError(quayConfiguration.QuayEcosystem, redhatcopv1alpha1.QuayEcosystemQuaySetupFailure, err)
 		}
 
-		err = r.quaySetupManager.SetupQuay(quaySetupInstance)
-
-		if err != nil {
-			r.reconcilerBase.GetRecorder().Event(quayConfiguration.QuayEcosystem, "Warning", "Failed to Setup Quay", err.Error())
-			return r.manageError(quayConfiguration.QuayEcosystem, redhatcopv1alpha1.QuayEcosystemQuaySetupFailure, err)
+		if !quayEcosystem.Spec.SkipValidation {
```
cnuland: Not sure if this make sense... it allows the unit test to validate the deployment of clair, but if the skipValidation flag is being used then it makes the pods that are being setup in setupQuay() dead in the water. I'd like to try to avoid flags used strictly for testing if possible. Thoughts @sabre1041 ? I'm thinking we just won't be able to test much of the reconcile cycle within the fake client and instead test it mostly within the E2E tests.

sabre1041: I would not recommend the `skipValidation` approach. Surely we should be able to use the fake client and related testing tools to implement this

pkg/controller/quayecosystem/resources/deployments.go
sabre1041: Is the value always `new-installation`?

pkg/k8sutils/k8sutils.go
sabre1041: This is good, I will go and incorporate the same into my [native k8s PR](https://github.com/redhat-cop/quay-operator/pull/125/files)

pkg/controller/quayecosystem/resources/deployments.go
cnuland: For any new instance of Quay, that is correct, the other possible options are, 

add-new-fields
backfill-then-read-only-new
remove-old-field

Considering what we're doing with the operator the new-installation makes the most sense since we will be deploying  a new instance of Quay, not modifying it.

pkg/controller/quayecosystem/quayecosystem_controller.go
cnuland: Removed skipValidation

pkg/controller/quayecosystem/provisioning/provision.go
sabre1041: Would you be able to elaborate? This logic would break the basics of operators to conform state. Please revert these changes

pkg/controller/quayecosystem/resources/deployments.go
sabre1041: A new field on the `quay` resource (enum type) should be created with the possible options (default to `new-installation`)

pkg/k8sutils/k8sutils.go
sabre1041: This does not work as the durations are in ms and fail immediately. Plus update accordingly

cnuland: updated

pkg/controller/quayecosystem/resources/deployments.go
cnuland: done

pkg/k8sutils/k8sutils.go
sabre1041: timeout duration is not correct. please change 42->420

pkg/k8sutils/k8sutils.go
```diff
@@ -29,55 +31,34 @@ func GetK8sClient(config *rest.Config) (kubernetes.Interface, error) {
...
+
+		logging.Log.Info("Waiting for deployment", "Namespace", namespace, "Name", name, "Timeout", timeout)
 		// check before watching in case the deployment is already scaled to 1
-		deployment, err := k8sclient.AppsV1().Deployments(namespace).Get(name, metav1.GetOptions{})
-		if err != nil {
-			logging.Log.Error(err, "Failed to get Deployment", "Name", deployment.Name)
-			return false
+		fieldSelector := fields.OneTermEqualSelector("metadata.name", name)
```
sabre1041: Logic is not correct. Polling never verifies successful deployment.

pkg/k8sutils/k8sutils.go
cnuland: Done, along with fixed a false positive with the polling loop.

sabre1041 (approved):
LGTM. Thanks @cnuland !

Issue #133, opened 2020-01-17
quay-operator crashes when adding rados registryBackends to CR


gorantornqvist:
When I try to add a rados registry backend to my CR:

```
apiVersion: redhatcop.redhat.io/v1alpha1
kind: QuayEcosystem
metadata:
  name: my-quayecosystem
spec:
  quay:
    registryBackends:
       - name: rados
         rados:
           hostname: rook-ceph-rgw-object-store.rook-ceph
           port: 8080
           secure: false
           accessKey: ********
           secretKey: ********
           bucketName: quay-bucket-6a744ef2-a516-4d09-b3b8-31f56db86146
```

The quay-operator crashes with error:

```
{"level":"info","ts":1579245717.8176355,"logger":"controller_quayecosystem","msg":"Reconciling QuayEcosystem","Request.Namespace":"quay-enterprise","Request.Name":"my-quayecosystem"}
E0117 07:21:57.918246       1 runtime.go:73] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1612 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x13dcda0, 0x219e050)
	/home/travis/gopath/pkg/mod/k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/pkg/util/runtime/runtime.go:69 +0x7b
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/home/travis/gopath/pkg/mod/k8s.io/apimachinery@v0.0.0-20190817020851-f2f3a405f61d/pkg/util/runtime/runtime.go:51 +0x82
panic(0x13dcda0, 0x219e050)
	/home/travis/.gimme/versions/go1.13.linux.amd64/src/runtime/panic.go:679 +0x1b2
github.com/redhat-cop/quay-operator/pkg/controller/quayecosystem/validation.Validate(0x17b4600, 0xc000351fb0, 0xc000460a80, 0x0, 0xc000304638, 0xc000bf9060)
	/home/travis/gopath/src/github.com/redhat-cop/quay-operator/pkg/controller/quayecosystem/validation/validate.go:358 +0x19d4
[...]

PR #136, opened 2020-01-22
Corrected RADOS storage issues in validation

sabre1041:
Corrected misconfiguration of variables for RADOS for #133


kurtismullins (approved):
LGTM

I am not entirely sure if this will solve the addressed issue because I do not have a ceph environment standing to verify against, but these changes look like they were required regardless.

PR #139, opened 2020-01-30
Updated permissions for quay operator binary

sabre1041:
Fix permissions for the quay-operator binary lost when uploading/downloading artifacts for #133


kurtismullins (approved):
Good catch! I'm surprised its losing the permissions bit. LGTM

PR #144, opened 2020-02-07
Added instructions for helm chart install from repo

sabre1041:
Added instructions for helm chart install from repo


kurtismullins (approved):
LGTM

PR #146, opened 2020-02-09
1.0.2 release

sabre1041:
1.0.2 release


deploy/olm-catalog/quay/1.0.2/quay.v1.0.2.clusterserviceversion.yaml
```diff
@@ -0,0 +1,370 @@
...
+              "configHostname": "example-quayecosystem-quay-enterprise.apps",
+              "configSecretName": "quay-config-app",
+              "database": {
+                "credentialsSecretName": "quay-database-credential",
+                "volumeSize": "10Gi"
+              },
+              "deploymentStrategy": "RollingUpdate",
+              "envVars": [
+                {
+                  "name": "DEBUGLOG",
```
kurtismullins, line 54: @sabre1041 It might be worth removing this environment variable.
The debug-level logging includes a *lot* of messages from Quay which aren't very important except for diagnostic purposes.

kurtismullins (approved):
After a quick visual inspection, it LGTM. Let me know your thoughts on changing the log-level.

deploy/olm-catalog/quay/1.0.2/quay.v1.0.2.clusterserviceversion.yaml
sabre1041: These are examples that can be used. I reordered them so that the simple example should be the one that is shown by default in OLM

PR #148, opened 2020-02-12
Modified QuayConfig route logic

sabre1041:
modified logic that manages creating the config quay route. will always be created whenever the pod is running


kurtismullins (approved):
LGTM

PR #159, opened 2020-03-03
GetDeploymentStatus support replica count greater than one

tjololo:
Update GetDeploymentStatus to check against .Status.Replicas

Check against UpdatedReplicas instead of AvailableReplicas

### DIFF
--- a/pkg/k8sutils/k8sutils.go
+++ b/pkg/k8sutils/k8sutils.go
@@ -53,22 +53,22 @@ func GetDeploymentStatus(k8sclient kubernetes.Interface, namespace string, name 
 		if !ok {
 			logging.Log.Error(err, "Unexpected type")
 		}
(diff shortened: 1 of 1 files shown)


sabre1041:
@tjololo thank you for this enhancement! However, would you be able to return the check against the `availableReplicas` instead of `updatedReplicas` as it will take the `minReadySeconds` value into consideration?

tjololo:
> @tjololo thank you for this enhancement! However, would you be able to return the check against the `availableReplicas` instead of `updatedReplicas` as it will take the `minReadySeconds` value into consideration?

Sure, made the changes

sabre1041 (approved):
LGTM. Thanks!

arve0:
@sabre1041 Any plans to release this to the stable channel?

PR #164, opened 2020-03-11
WIP: First batch of changes for ingress

sabre1041:
Large number of changes to support various ingress and termination methods


kurtismullins (approved):
LGTM

Let's get this merged in, start poking at it, work on getting this stabilized, and write any documentation on the upgrade path.

Nice work!

PR #168, opened 2020-03-18
PROJQUAY-479 - correct mirror probes

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-479

The livenessProbe and readinessProbe for mirror deployment should be the same as quay

### DIFF
--- a/pkg/controller/quayecosystem/resources/deployments.go
+++ b/pkg/controller/quayecosystem/resources/deployments.go
@@ -250,7 +250,7 @@ func GetQuayRepoMirrorDeploymentDefinition(meta metav1.ObjectMeta, quayConfigura
 				InitialDelaySeconds: 10,
 				Handler: corev1.Handler{
 					TCPSocket: &corev1.TCPSocketAction{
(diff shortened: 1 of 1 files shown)


kurtismullins:
@thomasmckay When the Quay worker is deployed as its own pod/container, does it provide a health check?

thomasmckay:
@kurtismullins not yet but this is the proper check for the operator

edit: maybe a different endpoint is better. in any case, someone review this to see if this is sufficient for when endpoint is determined

kurtismullins:
> @kurtismullins not yet but this is the proper check for the operator

Until that health-check is in place, I believe this will cause the deployment to fail. We might want to get that implemented in Quay before modifying it here.

sabre1041:
@thomasmckay Will not be able to accept this as the same health check that is used for Quay cannot be used for the repo mirror

The operator currently uses a TCP check to see whether port 9092 is open which should suffice

sabre1041:
@thomasmckay I think you can close this issue. It has been implemented as part of #187

PR #183, opened 2020-04-05
Disable secrets controller

sabre1041:
Disable secrets controller until we stabilize features


kurtismullins (approved):

kurtismullins:
Looks good to me while we stabilize other features. Let's try to make sure we can re-enable, or otherwise fix, this before the release.

Issue #175, opened 2020-04-03
Clair connecting to database with sslmode=disable


JuozasA:
When Clair 3.2 is being setup with Operator, clair-config.yaml has by default:
source: postgres://username:password@service/clair?sslmode=disable

https://github.com/redhat-cop/quay-operator/blob/62dc01ad2ce421f794e546e1cdf3c8283bab454c/pkg/controller/quayecosystem/provisioning/provision.go#L677

Would be useful either to have a flag in QuayEcosystem for ssl true/false or set it to sslmode=require by default

kurtismullins:
I agree that it should at least have the option to use an encrypted connection. I believe the Operator would have to generate a key for the Postgres instance and then supply that key to Clair and the Postgres container. I don't have much experience with Clair but I can reach out to some of the Clair maintainers for details if needed.

JuozasA:
Or maybe, if we specify our own DB in QuayEcosystem - enforce ssl (me personally was using Azure Postgresql and I don't really want to disable ssl there), if we are asking for operator to create a DB for us, disable ssl...

PR #184, opened 2020-04-05
connectionOptions setting for Clair in QuayEcosystem

JuozasA:
When Clair is being setup with Operator, clair-config ConfigMap has by default:
`source: postgres://username:password@service/clair?sslmode=disable`

https://github.com/redhat-cop/quay-operator/blob/62dc01ad2ce421f794e546e1cdf3c8283bab454c/pkg/controller/quayecosystem/provisioning/provision.go#L677

This PR is related to #175 

Proposed feature is to keep sslmode=disable if it is not set in QuayEcosystem. 
In order to set the different sslmode (accepted are disable, require, verify-ca or verify-full) clairDatabaseConSslMode key can be added to Clair property in QuayEcosystem CR:
```
<...>
spec:
  clair:
    enabled: true
    clairDatabaseConSslMode: require
    imagePullSecretName: redhat-pull-secret
<...>
```


sabre1041:
@JuozasA this is a good idea and thanks for the initial contribution. To allow for additional connection properties to be defined, how about we instead allow for a map of options in a field called `databaseConnectionOptions`?

Thoughts?

JuozasA:
> @JuozasA this is a good idea and thanks for the initial contribution. To allow for additional connection properties to be defined, how about we instead allow for a map of options in a field called `databaseConnectionOptions`?
> 
> Thoughts?

Just to check if I understood correctly, databaseConnectionOptions should be available for both, Quay and Clair with sslmode options or with all available database connection options?

sabre1041:
@JuozasA I would recommend only integrating with Quay and Clair for connectivity to PostgreSQL

JuozasA:
Quay is using sslmode=prefer by default (which works like this: _I don't care about encryption but will pay the overhead of encryption if the server supports it_). 
Clair does not accept `prefer` sslmode (only accepts `disable`, `require`, `verify-ca` or `verify-full`), so the mode needs to be set in the connector URI. 
Not sure we need to add databaseConnectionOptions for Quay since it will connect to both, ssl enabled and not enabled databases due to default sslmode being used

sabre1041:
@JuozasA lets at least add it for clair

JuozasA:
@sabre1041 added for Clair. Users can provide connection string parameters as key: value pairs in QuayEcosystem and they will be added to the connection string in ConfigMap. 
I've tried to add the same for Quay, however, since reconciliation is slightly different for quay database check, I need to go through the whole quay code part :)

sabre1041:
Receiving the following errors when generating client stubs 

```
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [redhatcop:[v1alpha1], ] 
F0421 19:04:31.424852   30331 deepcopy.go:750] DeepCopy of "interface{}" is unsupported. Instead, use named interfaces with DeepCopy<named-interface> as one of the methods.
```

I dont see those included in your PR

JuozasA:
> Receiving the following errors when generating client stubs
> 
> ```
> INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [redhatcop:[v1alpha1], ] 
> F0421 19:04:31.424852   30331 deepcopy.go:750] DeepCopy of "interface{}" is unsupported. Instead, use named interfaces with DeepCopy<named-interface> as one of the methods.
> ```
> 
> I dont see those included in your PR

completely forgot to run generate k8s... 🤦 
fixed

sabre1041 (approved):
LGTM.

Thanks @JuozasA for this.

I have one comment. Do we want to rename to `connectionParameters` since it is already within the `database` property and `databaseConnectionParameters` is somewhat redundant.

JuozasA:
> LGTM.
> 
> Thanks @JuozasA for this.
> 
> I have one comment. Do we want to rename to `connectionParameters` since it is already within the `database` property and `databaseConnectionParameters` is somewhat redundant.

Yes, good point.

sabre1041 (approved):
LGTM

Issue #179, opened 2020-04-03
Clair - There is no way to pass extra CA certificates to Clair pod


scouturier:
Currently, there is no documented ways to pass extra CA certificates to Clair pod, therefore if Quay is using a self-signed (with CA) certificate on it's route, Clair will not trust it unless the certificate is manually added to the Deployment.

Could be similar to the extraCaCert for quay:
```
  quay:
    configFiles:
      - secretName: quayconfigfile
        type: extraCaCert
```

liveaverage:
Just ran into the same thing and incorrectly assumed extraCaCert definitions for clair would work. Like you, manually added the CA bundle directly to the Deployment (mounting at `$CLAIRDIR/certs/` per https://github.com/quay/clair-jwt/blob/master/generate_mitm_ca.rhel.sh#L14) and everything worked fine... but it'd be nice to see this added to the operator.

sabre1041:
This feature is currently being worked on and should be available as part of the next release

sabre1041:
Resolved as part of #187

Issue #174, opened 2020-04-03
Operator is not Adding SUPER_USERS to Geo-replicated Quay instance


JuozasA:
I am using Operator 1.0.2 for both Quay 3.2 instances on separate OCP 4.3 clusters. 

1st Quay instance gets deployed successfully (when the database is empty), but when I use Operator for 2nd Quay deployment, it skips super user creation (because user already exists in the shared database) but with that, it is not adding SUPER_USERS field into Quay config.yaml so after deployment is finished, 2nd Quay instance does not have Super User Admin Panel in the GUI and also, clair security-scanner secret is not being created (even if clair is not sharing the database) on the 2nd instance. 

Would be good if operator would check if super user already exists in the database and populate those fields into config.yaml and create security scanner secret if one does not exist

ibazulic:
If you deploy Quay instances in two separate locations but you want them to work in georeplication mode, you have to have **the same** configuration on both places. Additionally:

- you need to have a unique entrypoint for all Quay instances/clusters for auth to work
- all instances need to have access to all storage engines
- all instances need to have access to the same database

All these requirements are satisfied **IF and only IF** setup is done on one location and then that config.yaml file is copied to all other locations. You cannot deploy Quay on the 2nd location from scratch.

JuozasA:
I understand that, I am fine with copying the quay and clair configs, security_scanner private key to another cluster, but I would still want to use Operator for deploying other components like mirror pod and clair, but right now, if I set skipSetup: true, operator stops on quay deployment and will not progress with clair, or mirror deployments

ibazulic:
Please see: 

https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/deploy_red_hat_quay_on_openshift_with_quay_setup_operator/customizing_your_red_hat_quay_cluster#injecting_configuration_files

ibazulic:
Create a secret called quayconfig:

~~~
oc create secret quayconfig --from-file=/path/to/quay/config.yaml
~~~

Then, define the config file as follows:

~~~
apiVersion: redhatcop.redhat.io/v1alpha1
kind: QuayEcosystem
metadata:
  name: example-quayecosystem
spec:
  quay:
    configFiles:
      - secretName: quayconfig
~~~

Finally, add the `skipSetup: true` at the end. It should pull the config.yaml from the secret and mount it to the correct place in the pod.

JuozasA:
[...]

Issue #165, opened 2020-03-13
Operator needs to clean up SCC Bindings when destroying a QuayEcosystem


wkulhanek:
When I did my tests on a new cluster everything worked every time. Until it failed every time.

I cleaned up the anyuid scc (removed all the service accounts that the quay operator added) and sure enough the next deploy succeeded again.

Now.... I only had 5 (!!!) service account pairs with the SCC. So we need to file a bugzilla as well for the product.

But the operator really needs to clean these up when deleting a QuayEcosystem.

wkulhanek:
Interestingly enough that did not work on RHPDS (OCP 4.1.30). It did work on DEV4 (OCP 4.3.3). And on RHPDS it's actually missing the quay SA from the anyuid SCC.

sabre1041:
This feature requires the use of a finalizer for the operator to act upon

Items to be addressed: 

1. Name of finalizer to be deployed

sabre1041:
Resolved as part of #187

Issue #149, opened 2020-02-14
Unable to Set User as Super User


KrisJohnstone:
All the documentation mentions that we are able to have LDAP integration with Quay, however when I try to follow the instructions located here:

https://access.redhat.com/documentation/en-us/red_hat_quay/2.9/html/manage_red_hat_quay/ldap-authentication-setup-for-quay-enterprise

I encounter an issue where I'm unable to elevate a user to Super User. Screen shot:
https://imgur.com/k5bw97F

Using OpenShift 4.3 and the Quay Operator 1.0.1. If modifying a users perms or LDAP isn't possible then should probably add a section about what features aren't implemented yet.

sabre1041:
Resolved as part of #187

PR #187, opened 2020-04-11
Changes/Enhancements leading up to 1.1.0 

### LINKED ISSUE #178: Certs uploaded through the Config app break the Quay pod.
**Description:**

If a User uploads a self-signed certificate to communicate with external services (e.g. LDAP) using the Quay Configuration application, then the Quay pods will fail on initilization.

The error indicates this is related to the mounted secret at `/conf/stack` being read-only, which is expected behavior when using a secret. However, this has worked in the past without the operator.

Here is an important portion of the logs as shared by Quay's QE:
```
Running all default registry services
Running init script '/quay-registry/conf/init/02_get_kube_certs.sh'
Running on kubernetes, attempting to retrieve extra certs from secret
Found an extra cert %s in config-secret, copying to kube ca dir
Traceback (most recent call last):
File "/quay-registry/conf/init/02_get_kube_certs.py", line 81, in <module>
main()
File "/quay-registry/conf/init/02_get_kube_certs.py", line 76, in main
with open(os.path.join(KUBE_EXTRA_CA_CERTDIR, cert_filename), "w") as f:
IOError: [Errno 30] Read-only file system: u'/conf/stack/extra_ca_certs/noobaas3.crt'
```

This is a reference to the associated file which attempts to fetch the "missing" certificate and store it locally: https://github.com/quay/quay/blob/master/conf/init/02_get_kube_certs.py

I believe the issue is related to the way the secret is mounted. Quay finds a certificate in its secret but does not see it locally, so it tries to write it locally.

**Steps to Reproduce**

1. Deploy Quay Ecosystem. Configure CR to *not* stop the Configuration Application.
2. Access the Configuration Application.
3. Use the "Upload Certifiates" to upload any valid certificate with the name `foo.crt`.
4. Save the configuration.
5. The file `foo.crt` should now exist in Quay's configuration secret.
6. Restart the Quay Pod(s)
7. They should fail with the error mentioned above.

**Acceptance Criteria**

Users should be able to upload certificates through the Quay Config application and any Quay pods should be able to use those certificates successfully.

sabre1041:
Numerous changes and enhancements to address issues and land new features in advance of release version 1.1.0

1. Internal restructuring of how `extraCaCerts` are managed
2. Fixes to SCC management
3. Implementation of Finalizers to aid in automatically clean up SCC's
4. Support for `configFiles` for Clair
5. Number of repo mirror replicas
6. Probes for repo mirror
7. Support for managing Superusers

Resolves #178 #135 #165 #179 #174 #149


README.md
```diff
@@ -555,6 +557,25 @@ spec:
...
+  name: example-quayecosystem
+spec:
+  quay:
+    imagePullSecretName: redhat-pull-secret
+    superusers:
+      - jim
+      - joe
+```
+
+If multiple superusers are specified, during the initial setup of Quay, the first user specified will be configured unless specified within a secret as described earlier. After the initial setup, passwords are managed witihn Quay itself and not using either the default value or the value provided in the secret.
```
kurtismullins, line 577: We should mention that User's default password here.

kurtismullins (commented):
Just a single comment so far. Will review one last time after it's changed from WIP

kurtismullins (approved):
LGTM

PR #195, opened 2020-04-22
Fix examples

scouturier:
Updating examples parameters to "hostname" to reflect documentation.


deploy/examples/005_advanced_config_example_cr.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     deploymentStrategy: RollingUpdate
     skipSetup: false
     externalAccess:
-      configHostname: example-quayecosystem-quay-enterprise.apps
+      hostname: example-quayecosystem-quay-enterprise.apps
```
sabre1041: Does this require a change? `configHostname` is correct for specifying the location for the Config Route

sabre1041 (commented):
@scouturier Thanks for helping update some of the examples. Added one comment

deploy/examples/005_advanced_config_example_cr.yaml
scouturier: You are right, I will fix that and resubmit!

sabre1041 (approved):
LGTM. Thanks!

PR #196, opened 2020-04-23
Add Quay Metrics Port 9091 to Quay deployment and service

JuozasA:
Since Prometheus Pushgateway is used for metrics and it can be accessed via port 9091, adding it to the Quay service.

P.S.
Exists in OCP templates as well:
https://github.com/quay/quay/blob/e919361a9aad73df015c692e31a1fe1d015cafa4/deploy/openshift/quay-app.yaml#L62


sabre1041:
@JuozasA still has 16 commits...

PR #203, opened 2020-04-26
Remove configNodePort from README as it will not be used in most cases by normal users

sabre1041:
The use of `configNodePort` is really on used when running and testing the operator outside of the cluster. Removed from `README` to avoid confusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,7 +408,7 @@ spec:
 
 #### NodePorts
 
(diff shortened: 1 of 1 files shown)


kurtismullins:
@sabre1041 

If Users are using `nodeport` for Quay, itself, would it be unlikely that they would also need to access the configuration application through a nodeport? I don't have experience with most on-premise Quay deployments or the typical use-case for exposing applications via node-ports.

kurtismullins:
@sabre1041 This PR has a conflict with `README.md`

PR #211, opened 2020-05-08
Use latest Project Quay images

kurtismullins:
The latest upstream Project Quay images from the sprint `padme` are now available on quay.io. This Pull Request uses those images.

### DIFF
--- a/pkg/controller/quayecosystem/constants/constants.go
+++ b/pkg/controller/quayecosystem/constants/constants.go
@@ -14,13 +14,13 @@ const (
 	// OperatorName is a operator name
 	OperatorName = "quay-operator"
 	// QuayImage is the Quay image
-	QuayImage = "quay.io/projectquay/quay:latest"
+	QuayImage = "quay.io/projectquay/quay:padme"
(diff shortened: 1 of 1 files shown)


sabre1041 (approved):
@kurtismullins Tested successfully. Only question I have is whether we want to remove the explicit inclusions of the `imagePullSecret` from the documentation for the upstream operator?

kurtismullins:
> @kurtismullins Tested successfully. Only question I have is whether we want to remove the explicit inclusions of the `imagePullSecret` from the documentation for the upstream operator?

Good question. I think it would be a good idea to make sure it's documented but it would be nice to simplify the sample CR and it might be more informative if we document it as its own setting.

I would back whatever decision you make here :)

sabre1041:
@kurtismullins i would suggest that the docs here align with the expected functionality by default. if images do not require auth by default, remove from the examples. Of course, we would still want to show how it can be configured as we already currently do in the section on how to access protected images

sabre1041 (approved):
Will be merging to keep moving things forward. Followup PR to address documentation needs

PR #212, opened 2020-05-09
Migrated directly modifying SCC's and use RBAC instead

sabre1041:
Removed directly modifying the anyuid SCC and removed the finalizer as it is no longer needed

Resolves #212


thomasmckay:
Will these changes _only_ work with OCP-4.5? Is it possible to if/else around things to maintain backward compatible?

sabre1041:
@thomasmckay functionality is currently available in OCP 4.3/4.4 (as well as potentially older releases)

kurtismullins:
@sabre1041 It looks like this PR may need a rebase.

sabre1041:
@kurtismullins done.

A new task must be added to the quay-docs side to remove the finalizer discussion

alecmerdler (approved):

PR #215, opened 2020-05-11
Removal of imagePullSecret from docs by default

sabre1041:
Removal of imagePullSecret from docs by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ $ helm install quay-operator quay-operator/quay-operator
 
 ### Deploy a Quay Ecosystem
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -112,8 +112,7 @@ kind: QuayEcosystem
 metadata:
   name: example-quayecosystem
 spec:
-  quay:
-    imagePullSecretName: redhat-pull-secret
+  quay: {}
```
thomasmckay, line 115: What do you think about adding a yaml comment at the above all of the `redhat-pull-secret` lines that says, "Required only if 'quay' and 'clair-jwt' images require pull secrets"

thomasmckay: alternatively, could put two example blocks, one with and one without the imagePullSecretName

sabre1041: im hesitant as that would really clutter up the documentation.

thomasmckay: How about two example blocks just here, one with and one without the secret. The rest merely omit.

thomasmckay (approved):
LGTM

PR #225, opened 2020-05-14
Documentation corrections for defining SSL certificates

### LINKED ISSUE #220: Improve error message when validating secrets to log the expected keys for the given secret.
When troubleshooting issues with secrets, the exception logs give a generic error message like:

```
{"level":"error","ts":1589377875.9836397,"logger":"controller_quayecosystem","msg":"Secret Validation","Namespace":"quay-enterprise","Name":"custom-quay-ssl","error":"Failed to validate provided secret with required parameters","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/tmp/quay-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/redhat-cop/quay-operator/pkg/controller/quayecosystem/validation.validateSecret\n\t/tmp/quay-
```

The problem in this case was that the docs say to use they key `tls.cert` for the ssl secret, but in fact the proper key is `tls.crt`. In order to figure that out, I had to trace all the way back through several file in the operator, which lead me back to the k8s core/v1 constants library to find the proper value (https://godoc.org/k8s.io/api/core/v1#pkg-constants). It would be create if the operator would just log the name of the keys epxected.

https://github.com/redhat-cop/quay-operator/blob/master/pkg/controller/quayecosystem/validation/validate.go#L512-L525

sabre1041:
Documentation fixes to resolve #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,18 +495,13 @@ Quay, as a secure registry, makes use of SSL certificates to secure communicatio
 
 SSL certificates can be provided and used instead of having the operator generate certificates. Certificates can be provided in a secret which is then referenced in the _QuayEcosystem_ custom resource. 
 
-The secret containing custom certificates must define the following keys:
-
-* `tls.cert` -  All of the certificates (root, intermediate, certificate) concatinated into a single file
-* `tls.key` - Private key as for the SSL certificate
-
 Create a secret containing the certificate and private key
 
 ```
-oc create secret tls custom-quay-ssl --key=tls.key=<ssl_private_key> --cert=tls.cert=<ssl_certificate>
+oc create secret tls custom-quay-ssl --key=<ssl_private_key> --cert<ssl_certificate>
 ```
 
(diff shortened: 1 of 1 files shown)


thomasmckay:
@sabre1041 As I suggested in this PR/223, it will be better for us to be working in the shipped docs (this same doc source is used for upstream and downstream).
https://github.com/quay/quay-docs/blob/660704578d45359d48641606731e6b976dbda6eb/modules/proc_deploy-quay-openshift-operator.adoc

chrisnegus:
I'm following Andrew's changes closely here. For now, I have this PR and a good set of diffs to work with for the other changes. The Red Hat Quay/Project Quay docs should be up to date with all the quay-operator docs by the end of today. I agree, however, that in the long term it would be best to somehow pool our efforts and maybe choose one source to work from.

sabre1041:
Closing as docs have moved to [quay-docs](https://github.com/quay/quay-docs)

PR #226, opened 2020-05-19
Ensure Extra Certs Uploaded Using Config App

alecmerdler:
### Description

Any secrets of `type: extraCaCert` in the `QuayEcosystem` resource will be uploaded to the config app API (`POST /api/v1/superuser/customcerts/<certname>`), which has the side effect of[ installing them into the config app container](https://github.com/quay/quay/pull/276). Then, when the Operator goes to validate (which uses the config app API) a database, registry, or any other connection that uses self-signed cert, the validation will not fail SSL requirements because the cert has been installed in the container.

Fixes https://issues.redhat.com/browse/PROJQUAY-461

### DIFF
--- a/pkg/client/quay.go
+++ b/pkg/client/quay.go
@@ -195,6 +195,22 @@ func (c *QuayClient) UploadFileResource(fileName string, content []byte) (*http.
 	return resp, quayStatusResponse, err
 }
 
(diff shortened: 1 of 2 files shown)


alecmerdler:
@sabre1041 Added `return fmt.Errorf()`'s if uploading cert fails.

alecmerdler:
Closing in favor of https://github.com/redhat-cop/quay-operator/pull/239

Issue #156, opened 2020-02-28
RHOCS configuration for Noobaa raising 'Missing required parameter(s) for storage rhocs'


rgordill:
Operator version 1.0.2.

Configuring Noobaa for Storage Backend:

```
...
    registryBackends:
    - name: rhocs
      rhocs:
        hostname: s3.noobaa.svc
        port: 80
        secure: false
        accessKey: 'BHlTdbuUeoxMnjkoOGUr'
        secretKey: '/oM3oPf1pXc5gLVh4KUeuhey7ujHFl4D+ldx0oz9'
        bucketName: quay
        storagePath: '/datastorage/registry' 
...
```

Checking the CRD immediately to know if everything is ok:

```
...
    registryBackends:
    - name: rhocs
      rhocs:
        accessKey: BHlTdbuUeoxMnjkoOGUr
        bucketName: quay
        hostname: s3.noobaa.svc
        port: 80
        secretKey: /oM3oPf1pXc5gLVh4KUeuhey7ujHFl4D+ldx0oz9
        storagePath: /datastorage/registry
    repoMirrorResources: {}
...
```
So it seems "secure" has not been saved in the CR. After that, a configuration exception is raised:

  - lastTransitionTime: "2020-02-27T23:05:59Z"
    lastUpdateTime: "2020-02-27T23:10:29Z"
    message: 'Failed to Validate Component: registry-storage Validation Failed: Missing
      required parameter(s) for storage rhocs'
    reason: ProcessingError
    status: "False"
    type: QuaySetupFailure

With secure: true, it is stored,

rgordill:
Seems the same issue as the first comment in #109

alecmerdler:
Resolved in #239

PR #227, opened 2020-05-21
Removed omitempty from is_secure

sabre1041:
Removed omitempty from is_secure registry backends


alecmerdler (approved):
I don't understand why this is needed, but okay.

alecmerdler:
Resolves #156

Issue #219, opened 2020-05-12
Provide ability to define StorageClass for Databases


sabre1041:
Currently, it is impossible to define a `StorageClass` for all databases that are generated by the operator.

sabre1041:
Resolved by #229

PR #229, opened 2020-05-25
#219 Provide ability to define StorageClass for Databases

drichtarik:
Not sure about constants.go, defaults.go or resource/types.go edits
deyployments.go has comment section that I saved if it proves to be somewhat right approach

### DIFF
--- a/pkg/controller/quayecosystem/provisioning/provision.go
+++ b/pkg/controller/quayecosystem/provisioning/provision.go
@@ -386,7 +386,7 @@ func (r *ReconcileQuayEcosystemConfiguration) createQuayDatabase(meta metav1.Obj
 
 	// Create PVC
 	if !utils.IsZeroOfUnderlyingType(r.quayConfiguration.QuayEcosystem.Spec.Quay.Database.VolumeSize) {
-		databasePvc := resources.GetDatabasePVCDefinition(meta, r.quayConfiguration.QuayEcosystem.Spec.Quay.Database.VolumeSize)
+		databasePvc := resources.GetDatabasePVCDefinition(meta, r.quayConfiguration.QuayEcosystem.Spec.Quay.Database.VolumeSize, &r.quayConfiguration.QuayEcosystem.Spec.Quay.Database.StorageClass)
 		databaseResources = append(databaseResources, databasePvc)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/controller/quayecosystem/constants/constants.go
```diff
@@ -112,6 +112,7 @@ const (
 	QuayRegistryStoragePath = "/datastorage/registry"
 	// QuayRegistryStoragePersistentVolumeStoreSize represents the size of the PersistentVolume that should be used for registry storage
 	QuayRegistryStoragePersistentVolumeStoreSize = "20Gi"
+	PersistentVolumeStorageClassName = "storageClassName"
```
sabre1041: No need for this constant value as no default will be used

pkg/controller/quayecosystem/provisioning/provision.go
```diff
@@ -429,7 +429,7 @@ func (r *ReconcileQuayEcosystemConfiguration) createClairDatabase(meta metav1.Ob
 
 	// Create PVC
 	if !utils.IsZeroOfUnderlyingType(r.quayConfiguration.QuayEcosystem.Spec.Clair.Database.VolumeSize) {
-		databasePvc := resources.GetDatabasePVCDefinition(meta, r.quayConfiguration.QuayEcosystem.Spec.Clair.Database.VolumeSize)
+		databasePvc := resources.GetDatabasePVCDefinition(meta, r.quayConfiguration.QuayEcosystem.Spec.Clair.Database.VolumeSize, &r.quayConfiguration.QuayEcosystem.Spec.Quay.Database.StorageClass)
```
sabre1041: Should be Clair's database reference and not Quay

pkg/controller/quayecosystem/resources/deployments.go
```diff
@@ -710,6 +710,15 @@ func GetDatabaseDeploymentDefinition(meta metav1.ObjectMeta, quayConfiguration *
 		})
 	}
 
+	/*	if !utils.IsZeroOfUnderlyingType(database.StorageClass) {
```
sabre1041: Can probably remove this snippet

pkg/controller/quayecosystem/resources/types.go
```diff
@@ -72,6 +72,7 @@ type DatabaseConfig struct {
 	Memory              string
 	VolumeSize          string
 	CredentialsName     string
+	StorageClass        string
```
sabre1041: This value is not needed

pkg/controller/quayecosystem/resources/storage.go
```diff
@@ -50,6 +50,7 @@ func GetDatabasePVCDefinition(meta metav1.ObjectMeta, volumeSize string) *corev1
 					corev1.ResourceName(corev1.ResourceStorage): resource.MustParse(volumeSize),
 				},
 			},
+			StorageClassName: storageClass,
 		},
```
sabre1041, line 54: Take a look at the method above for how the StorageClass is being added to the object. It should only be added if the value is defined and is not an empty string

sabre1041 (changes requested):
Looking good.  Made a few recommendations on some changes for the next draft iteration

drichtarik:
Thank you for the input and review. Updated. Hopefully I didn't miss anything.

sabre1041:
@drichtarik this looks very good!

sabre1041 (changes requested):
Can you generate the CRD's (4.x and 3.x) along with the generated classes + openapi specs. otherwise we should be good!

sabre1041 (approved):
LGTM

PR #238, opened 2020-06-03
Support for defining SecurityContext

sabre1041:
Support for defining SecurityContext


alecmerdler (approved):

sabre1041:
@alecmerdler I inadvertently used the `securityContext` on the container level instead of the pod level. Pushed new commit and tested successfully

alecmerdler (approved):

PR #258, opened 2020-07-10
Adds support for server in Postgres secret

### LINKED ISSUE #249: Postgres Server Attribute in Secret
Currently passing in an existing postgres instance for the quay operator requires a secret that contains the username, password, database name. A separate key in the YAML for QuayEcosystem claim is required to pass in the server. It seems more straight forward to accept the server in the secret as well. We can also keep this secret optional in order to maintain backward compatibility. 

As a note, I am open to taking on this issue as well.

krishchow:
This PR fixes #249.

## Proposed Changes
- Adds in support an optional key `server` inside of the Postgres. This key overrides the `server` key from the CRD spec if both are provided. This change supports has backward compatibility, and supports the server being provided inside of the CRD as well. 

It is well known that hardcoded credentials can open up security vulnerabilities, although the quay operator currently supports the username, password, database, etc. being stored in the secret, it omits the server's endpoint. This is still a piece of credential information, and as such, it should be treated as a secret.

### DIFF
--- a/pkg/controller/quayecosystem/constants/constants.go
+++ b/pkg/controller/quayecosystem/constants/constants.go
@@ -80,6 +80,8 @@ const (
 	DatabaseCredentialsDatabaseKey = "database-name"
 	// DatabaseCredentialsRootPasswordKey represents the key for the database root password
 	DatabaseCredentialsRootPasswordKey = "database-root-password"
+	// DatabaseCredentialsServer represents the key for the database server
+	DatabaseCredentialsServerKey = "database-server"
 	// QuayDatabaseCredentialsDefaultUsername represents the default database username
 	QuayDatabaseCredentialsDefaultUsername = "quay"
 	// QuayDatabaseCredentialsDefaultPassword represents the default database password
--- a/pkg/controller/quayecosystem/validation/validate.go
+++ b/pkg/controller/quayecosystem/validation/validate.go
@@ -146,6 +146,15 @@ func Validate(client client.Client, quayConfiguration *resources.QuayConfigurati
 			quayConfiguration.QuayDatabase.Password = string(databaseSecret.Data[constants.DatabaseCredentialsPasswordKey])
 			quayConfiguration.QuayDatabase.Database = string(databaseSecret.Data[constants.DatabaseCredentialsDatabaseKey])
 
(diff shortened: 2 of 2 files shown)


screeley44:
@sabre1041 can you have someone take a look at this proposed change that we discussed. thanks!

sabre1041 (changes requested):
@krishchow A few comments:

1. There appears to be no consideration on if `server` is defined in a secret and whether to provision an instance [link](https://github.com/redhat-cop/quay-operator/blob/master/pkg/controller/quayecosystem/provisioning/provision.go#L96)
2. There appears to be no consideration for Clair database as well

krishchow:
Thanks for the review @sabre1041 !

I have resolved your second point, and I have added similar support for Clair.

Can you please expand a little further on your first point? The linked conditional is inside of provison.go, which I believe runs after the validation stage. I am currently populating the `.Spec.Quay.Database.Server` during the validation process for the secret, which would mean that block does not execute.

sabre1041:
@krishchow would you be able to rebase the PR?

krishchow:
@sabre1041 rebased, let me know if this looks ok now

pkg/controller/quayecosystem/constants/constants.go
```diff
@@ -80,6 +80,8 @@ const (
 	DatabaseCredentialsDatabaseKey = "database-name"
 	// DatabaseCredentialsRootPasswordKey represents the key for the database root password
 	DatabaseCredentialsRootPasswordKey = "database-root-password"
+	// DatabaseCredentialsServer represents the key for the database server
+	DatabaseCredentialsServerKey = "server"
```
sabre1041: Can the key be called `database-server` to align with the rest of the database parameters?

sabre1041 (changes requested):
@krishchow Small change. Then will run through one final review

pkg/controller/quayecosystem/constants/constants.go
krishchow: updated to database-server

krishchow:
@sabre1041 I believe I have made all the changes you requested now, can we begin the final review?

screeley44:
@sabre1041 - any status on this, has the larger team had a chance to review or think we can get this one moving?

sabre1041 (approved):
LGTM

PR #315, opened 2020-09-16
Route Managed Component Enhancements (PROJQUAY-1091)

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1091

**Changelog:** Allow user-provided hostname and SSL certificate/key pair for `route` managed component.

**Docs:** There are now two scenarios for using the `route` managed component.
1. User creates and references `configBundleSecret` which contains `SERVER_HOSTNAME` field and `ssl.key` + `ssl.cert`. The created `Route` will have `passthrough` TLS termination and the custom hostname set. User will then need to create a CNAME record for their domain pointing to the router.
2. User creates zero-config `QuayRegistry`. A `Route` is created with a generated hostname and `edge` TLS termination using the cluster CA certs.

**Testing:** Create and reference a `configBundleSecret` which contains `SERVER_HOSTNAME` field and `ssl.key` + `ssl.cert` which are valid for that hostname. Check that the `Route` has the correct hostname and that the keypair is mounted into the Quay app `Deployment`.

**Details:** Previously, the `route` managed component used TLS edge-termination with the cluster certs and just targeted the `HTTP` port on the Quay pod, and also the default generated `spec.hostname`. Now we use TLS `passthrough` with the user-provided certificate/keypair and hostname if they have been provided in the config bundle. If not present, we fall back to existing behavior using cluster certificate and generated hostname.

### Screenshots

**Config editor UI showing managed `route` component w/ user-defined hostname**:
Screenshot_20200916_130546

### DIFF
--- a/controllers/quayregistry_controller.go
+++ b/controllers/quayregistry_controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-logr/logr"
 	objectbucket "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1"
 	routev1 "github.com/openshift/api/route/v1"
+	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


kustomize/components/route/route.varreference.yaml
```diff
@@ -0,0 +1,3 @@
+varReference:
```
alecmerdler: Adding a custom `varReference` following [this inspiration](https://github.com/kubernetes-sigs/kustomize/issues/1938#issuecomment-567974171).

PR #348, opened 2020-11-03
PROJQUAY-1236 - default DEBUGLOG false

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1236

**Changelog:** 
none

**Docs:** 
none

**Testing:** 
none

**Details:** 
DEBUGLOG for quay containers should be false. Running the config-tool is the exception.

### DIFF
--- a/kustomize/base/upgrade.deployment.yaml
+++ b/kustomize/base/upgrade.deployment.yaml
@@ -33,7 +33,7 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: DEBUGLOG
-              value: "true"
+              value: "false"
             - name: WORKER_COUNT_WEB
               value: "4"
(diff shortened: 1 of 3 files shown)


kustomize/base/upgrade.deployment.yaml
```diff
@@ -32,8 +32,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-            - name: DEBUGLOG
-              value: "true"
+            - name: `DEBUGLOG
```
alecmerdler: Fix this extra tick character.

thomasmckay: done, thanks

alecmerdler (approved):

PR #377, opened 2020-12-15
PROJQUAY-1177 - updates for downstream build

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1177

**Details:** 
Updating downstream build details


deploy/manifests/quay-operator/3.4.0/quay-operator.clusterserviceversion.yaml
```diff
@@ -39,7 +40,7 @@ spec:
   customresourcedefinitions:
     owned:
     - description: Represents a full Quay registry installation.
-      displayName: Quay Registry
```
alecmerdler: The display names of CRDs should not change between upstream/downstream. The `kind` of the CRD is `QuayRegistry`, so it's display name should be "Quay Registry".

alecmerdler (approved):

PR #396, opened 2021-02-11
PROJQUAY-1434 add OCP infra node tolerations

dmesser:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1434

**Changelog:** 

adds the required tolerations for Quay to be schedulable on OCP Infra nodes as per https://docs.openshift.com/container-platform/4.6/machine_management/creating-infrastructure-machinesets.html#assigning-machineset-resources-to-infra-nodes

**Docs:** 

tbd

**Testing:** 

tbd

**Details:** 

https://issues.redhat.com/browse/PROJQUAY-1566

### DIFF
--- a/kustomize/base/config.deployment.yaml
+++ b/kustomize/base/config.deployment.yaml
@@ -14,6 +14,10 @@ spec:
       labels:
         quay-component: quay-config-editor
     spec:
+      tolerations:
+        - effect: NoSchedule 
+          key: node-role.kubernetes.io/infra 
+          operator: Exists
       volumes:
         - name: config-bundle
           secret:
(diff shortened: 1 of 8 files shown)


dmesser:
Actually, we don't need this. In modern Kubernetes version there are admission controllers that allow to automatically apply tolerations to all pods in a namespace via a namespace-level annotation.

Something like this should suffice:

```
metadata:
  annotations:
    scheduler.alpha.kubernetes.io/defaultTolerations: >-
      [{"operator": "Exists", "effect": "NoSchedule", "key":
      "node-role.kubernetes.io/infra"}]
```

PR #402, opened 2021-02-25
PROJQUAY-1653 - downstream branding

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1653

**Details:** 
Set the name and logo to downstream default values.


alecmerdler (approved):
In the future we'll implement this as an environment variable, but this works for now.

PR #405, opened 2021-03-17
PROJQUAY-880: Add finalizer for cleanup operations

syed:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-880
Pull-request title must start with "PROJQUAY-880 - "

**Changelog:** 

Added a finalizer for running cleanup operations on deletion of a `QuayRegistry`

**Docs:** 

**Testing:** 

**Details:** 

The finalizer gets invoked when a `QuayRegsitry` resource is deleted. The finalizer should ideally be idempotent as it may be invoked multiple times.


controllers/quay/quayregistry_controller.go
```diff
@@ -84,6 +87,22 @@ func (r *QuayRegistryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	updatedQuay := quay.DeepCopy()
 	quayContext := quaycontext.NewQuayRegistryContext()
 
+	isQuayMarkedToBeDeleted := quay.GetDeletionTimestamp() != nil
+	if isQuayMarkedToBeDeleted {
+		if controllerutil.ContainsFinalizer(updatedQuay, QuayOperatorFinalizer) {
+			if err := r.finalizeQuay(ctx, updatedQuay); err != nil {
+				return ctrl.Result{}, err
```
alecmerdler, line 94: In the future, if the finalizer fails we might want to emit an event and set  a `status.condition` to convey the error to the user. But this is fine for now.

alecmerdler (approved):

PR #406, opened 2021-03-18
Remove Need for SANs in Provided Cert/Key Pair

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1737

**Changelog:** Remove the need for SANs of internal k8s service hosts in provided cert/key pair for Quay.

**Docs:** https://github.com/quay/quay-docs/pull/157

**Testing:** There are a few different situations which must be tested:

`Route` | `SERVER_HOSTNAME` | Custom cert/key pair | Expected result
|---|---|---|---|
| Managed | None | None | Uses generated `Route` hostname, self-signed certs |
| Managed | None | Provided (for generated `Route` hostname) | Uses generated `Route` hostname and provided certs |
| Managed | None | Provided (for random hostname) | Blocked rollout, error in `status.conditions` |
| Managed | `registry.company.com` | Provided (invalid) | Blocked rollout, error in `status.conditions` |
| Managed | `registry.company.com` | None | Uses provided hostname and self-signed certs |
| Managed | `registry.company.com` | Provided (valid) | Uses provided hostname and certs |
| Unmanaged | None | None | Error because `SERVER_HOSTNAME` is unset |
| Unmanaged | None | Provided (for random hostname) | Error because `SERVER_HOSTNAME` is unset |
| Unmanaged | `registry.company.com` | None | Uses provided hostname and self-signed certs |
| Unmanaged | `registry.company.com` | Provided (invalid) | Blocked rollout, error in `status.conditions` |
| Unmanaged | `registry.company.com` | Provided (valid) | Uses provided hostname and provided certs |

**Details:** Most public key infrastructure cannot generate cert/key pairs for the internal Kubernetes `Service` hostnames (`quay.namespace.svc`). Previously, the Operator would reject provided certs which did not include these hostnames as Subject Alternative Names (SANs) because they were believed to be needed for TLS connections within the cluster. This is not the case, so this check is removed.

To use a `Route` with the included OpenShift cluster cert/key pair, mark the `route` component as `managed: false`, include `SERVER_HOSTNAME` as the future generated `Route` hostname (follows the pattern `<route-name>-<namespace>.apps.<cluster>`), then create your own `Route` which points to the Quay app `Service` and uses re-encrypt TLS termination, with the `destinationCACertificate` set to the generated `tls.cert` found in the `Secret` mounted into the Quay app pods.

### DIFF
--- a/kustomize/base/quay.service.yaml
+++ b/kustomize/base/quay.service.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     quay-component: quay-app
 spec:
-  type: LoadBalancer
+  type: ClusterIP
   ports:
     - protocol: TCP
       name: https
--- a/pkg/kustomize/secrets_test.go
+++ b/pkg/kustomize/secrets_test.go
@@ -1,9 +1,8 @@
 package kustomize
 
 import (
-	"net"
+	"fmt"
 	"net/url"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -192,6 +191,7 @@ func TestFieldGroupFor(t *testing.T) {
 var containsComponentConfigTests = []struct {
 	name          string
 	component     v1.ComponentKind
+	managed       bool
 	rawConfig     string
 	expected      bool
 	expectedError error
@@ -199,6 +199,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"ClairContains",
 		"clair",
+		true,
 		`FEATURE_SECURITY_SCANNER: true`,
 		true,
 		nil,
@@ -206,6 +207,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"ClairDoesNotContain",
 		"clair",
+		true,
 		``,
 		false,
 		nil,
@@ -213,6 +215,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"PostgresContains",
 		"postgres",
+		true,
 		`DB_URI: postgresql://test-quay-database:postgres@test-quay-database:5432/test-quay-database`,
 		true,
 		nil,
@@ -220,6 +223,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"PostgresDoesNotContain",
 		"postgres",
+		true,
 		``,
 		false,
 		nil,
@@ -227,6 +231,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"RedisContains",
 		"redis",
+		true,
 		`BUILDLOGS_REDIS:
   host: test-quay-redis
 `,
@@ -236,6 +241,7 @@ var containsComponentConfigTests = []struct {
 	{
 		"RedisDoesNotContain",
 		"redis",
+		true,
 		``,
 		false,
 		nil,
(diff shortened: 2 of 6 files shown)


kustomize/base/quay.service.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   labels:
     quay-component: quay-app
 spec:
-  type: LoadBalancer
+  type: ClusterIP
```
dmesser, line 8: Can this change make a z-stream update fail?

alecmerdler: Potentially, I'll check and make sure. We already have a utility function which re-creates objects with immutable fields (we use for `Jobs`), so it may automatically do the right thing.

pkg/kustomize/secrets_test.go
```diff
@@ -319,94 +317,146 @@ func TestContainsComponentConfig(t *testing.T) {
...
-		if err != nil {
-			t.Errorf(err.Error())
-		}
-		assert.Equal(recPublicKey, pubWithBuilder)
-		assert.Equal(recPrivateKey, privWithBuilder)
-	})
+		assert.Equal(test.expectedErr, err, test.name)
 
+		if test.providedCert == nil || test.providedKey == nil {
+			shared.ValidateCertPairWithHostname(tlsCert, tlsKey, test.serverHostname, fieldGroupNameFor("rout"))
```
michaelryanmcneill: Is this supposed to be "rout" and not "route"?

alecmerdler: Yep

pkg/kustomize/kustomize.go
```diff
@@ -462,6 +462,7 @@ func Inflate(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, baseCo
 		}
 	}
 
+	// FIXME(alecmerdler): Should we create certs if `route` component is unmanaged...?
 	log.Info("Ensuring `ssl.cert` and `ssl.key` pair for Quay app TLS")
 	tlsCert, tlsKey, err := EnsureTLSFor(ctx, quay, baseConfigBundle.Data["ssl.cert"], baseConfigBundle.Data["ssl.key"])
```
alecmerdler: @dmesser Do you have opinions on whether we should generate self-signed certs for Quay when the `route` component is unmanaged?

alecmerdler: Having TLS termination at the Quay level (no edge termination) is much preferred.

dmesser: Unless the user provides customs certs thats ok. However in order for integration with Ingress and Routes to work we should expose these generated certs as `Secrets` (and link them on the CR) so users have a chance to set up a re-encrypting `Routes`: https://docs.openshift.com/container-platform/4.6/networking/routes/secured-routes.html#nw-ingress-creating-a-reencrypt-route-with-a-custom-certificate_secured-routes

alecmerdler: Yes, the suggestion will be to grab the self-signed certs from the Quay config bundle `Secret` and create a re-encrypt `Route`. That should satisfy everyone.

PR #420, opened 2021-03-30
Remove Owner References from Grafana ConfigMap (PROJQUAY-1788)

jonathankingfc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1788

**Changelog:** 
- Reconcile loop ensures that no `ownerReference` are injected on the grafana-dashboard config map. This prevents the cross-namespace owner references that caused resource to be garbage collected after creation.

**Docs:** 
- N/A

**Testing:** 
- Install operator and create instance with monitoring enabled
- ConfigMap <operator-namespace>-grafana-dashboard-quay should be visible in openshift-config-managed namespace

**Details:** 
- N/A
------

### DIFF
--- a/apis/quay/v1/quayregistry_types.go
+++ b/apis/quay/v1/quayregistry_types.go
@@ -357,6 +357,25 @@ func EnsureOwnerReference(quay *QuayRegistry, obj runtime.Object) (runtime.Objec
 	return obj, nil
 }
 
+// RemoveOwnerReference removes the `ownerReference` of `QuayRegistry` on the given object.
+func RemoveOwnerReference(quay *QuayRegistry, obj runtime.Object) (runtime.Object, error) {
+	filteredOwnerReferences := []metav1.OwnerReference{}
+
+	objectMeta, err := meta.Accessor(obj)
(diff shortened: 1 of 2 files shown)


alecmerdler (changes requested):
Add finalizer logic to replace garbage collection, and change `RemoveOwnerReference` to accept the `metadata.name` of a `QuayRegistry` to remove its specific `ownerReference` from the list.

apis/quay/v1/quayregistry_types.go
```diff
@@ -357,6 +357,25 @@ func EnsureOwnerReference(quay *QuayRegistry, obj runtime.Object) (runtime.Objec
 	return obj, nil
 }
 
+// RemoveOwnerReference removes the `ownerReference` of `QuayRegistry` on the given object.
+func RemoveOwnerReference(quay *QuayRegistry, obj runtime.Object) (runtime.Object, error) {
```
alecmerdler, line 361: Could you add a simple test for this?

alecmerdler (approved):

PR #422, opened 2021-03-31
Ensure Correct Clair Config Secret is Used (PROJQUAY-1797)

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1797

**Changelog:** Add suffix hash to generated Clair config `Secret` to ensure the correct version is mounted into the pod on re-reconcile.

**Docs:** N/a

**Testing:** N/a

**Details:** Previously had `disableNameSuffixHash: true` on the Kustomize-generated Clair config `Secret`. This meant that the `Secret` was updated and the name stayed the same, causing a race condition between the Clair pod starting up and the contents of the `Secret` being updated (with the new PSK) after a reconcile.


jonathankingfc (approved):
LGTM

PR #425, opened 2021-04-09
ServiceAccounts for Managed Postgres Components 

alecmerdler:
Use a separate `ServiceAccount` for each managed Postgres `Deployment` (rather than the `default` one in the namespace). This is needed when the Quay app pod needs to run with `anyuid` SCC (for `quayio` branch), which breaks the permissions on Postgres container for mounting the `pgdata` directory.


jonathankingfc (dismissed):
LGTM

jonathankingfc (approved):
Approved

PR #426, opened 2021-04-10
Python2 Images (quayio)

alecmerdler:
Update the Quay app container images to use the most recent Python2 build from the `quayio` branch.


alecmerdler:
Closing as it seems that you can run the Operator as-is and just use image overrides (`RELATED_IMAGE_COMPONENT_QUAY`).

Issue #400, opened 2021-02-19
Role "skynet-quay-database" does not exist


djjudas21:
I'm new to Quay but I've followed the "batteries-included, zero-config" instructions in the README on my OKD cluster. It's unable to start up because of a problem with Postgres that causes it to crashloop:

Unhealthy deployment:

```
[jgazeley@hephaestos quay-operator]$ oc get po
NAME                                         READY     STATUS                  RESTARTS   AGE
quay-operator-tng-68fd4fcbc8-r46cd           1/1       Running                 0          19m
skynet-quay-app-upgrade-6944487677-7tcbw     0/1       CrashLoopBackOff        8          18m
skynet-quay-config-editor-596bf4dbdc-rmdp6   1/1       Running                 0          18m
skynet-quay-database-6cfd8cb54-sshfr         0/1       Running                 2          4m23s
skynet-quay-mirror-58567f9b6-59h4v           0/1       Init:CrashLoopBackOff   5          17m
skynet-quay-postgres-init-jjbpl              0/1       Completed               0          18m
skynet-quay-postgres-init-q2p8w              0/1       Completed               0          18m
skynet-quay-redis-bfcb7995b-lqz5d            1/1       Running                 0          18m
```

Postgres having problems:

```
[jgazeley@hephaestos quay-operator]$ oc logs skynet-quay-database-6cfd8cb54-sshfr
pg_ctl: another server might be running; trying to start server anyway
waiting for server to start....2021-02-19 13:33:25.395 UTC [25] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2021-02-19 13:33:25.435 UTC [25] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
2021-02-19 13:33:26.086 UTC [25] LOG:  redirecting log output to logging collector process
2021-02-19 13:33:26.086 UTC [25] HINT:  Future log output will appear in directory "log".
........................................................................ done
server started
/var/run/postgresql:5432 - accepting connections
=> sourcing /usr/share/container-scripts/postgresql/start/set_passwords.sh ...
ERROR:  role "skynet-quay-database" does not exist
```

And of course, stuff that depends on Postgres is broken too:

```
[jgazeley@hephaestos quay-operator]$ oc logs skynet-quay-app-upgrade-6944487677-7tcbw
   __   __
  /  \ /  \     ______   _    _     __   __   __
[...]

PR #427, opened 2021-04-12
Fix Managed Database Initialization (PROJQUAY-1664)

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1664

**Changelog:** Fixes `role does not exist` error for managed database component by adding `startupProbe`.

**Docs:** N/a

**Testing:** Create `QuayRegistry` with managed database and ensure all health checks pass.

**Details:** Fixes an issue where the managed Postgres pod is killed before it completes initialization, and the next pod that starts up reports `role does not exist` error. Followed the [recommendations of the Postgres container maintainers](https://github.com/sclorg/postgresql-container/issues/309) and added `startupProbe` which tells k8s when the container has finished initialization. Apparently the `livenessProbe` [always returns true](https://github.com/sclorg/postgresql-container/pull/320).

### DIFF
--- a/kustomize/base/quay.deployment.yaml
+++ b/kustomize/base/quay.deployment.yaml
@@ -56,17 +56,24 @@ spec:
             limits:
               cpu: 2000m
               memory: 8Gi
-          readinessProbe:
-            exec:
-              command:
-                - curl
-                - '-k'
-                - 'https://localhost:8443/health/instance'
-            initialDelaySeconds: 30
+          startupProbe:
+            httpGet:
+              path: /health/instance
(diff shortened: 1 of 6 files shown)


alecmerdler:
Fixes #400

jonathankingfc:
Tested this PR locally and while the Postgres issue appears to be fixed, the `quay-app-upgrade` pod is never replaced with the `quay-app` pod after migrations have finished. The final state of reconciliation leaves the `quay-app-upgrade` pod running. The `quay-app` pod is never brought up.

PR #435, opened 2021-04-15
[WIP] Admission Webhooks (PROJQUAY-1622)

alecmerdler:
webhook: add k8s admission webhooks (PROJQUAY-1622)

Add a mutating and validating admission webhook to
fulfill the tasks of populating default values on
a QuayRegistry, and validating it and the referenced
'spec.configBundleSecret'.


ricardomaraschini:
Closing this thingy, if we need this we will implement in a new PR as this is already pretty much incompatible with what we have.

PR #447, opened 2021-04-22
clair: point liveness probe at introspection server

hdonnay:

### DIFF
--- a/kustomize/components/clair/clair.deployment.yaml
+++ b/kustomize/components/clair/clair.deployment.yaml
@@ -38,15 +38,16 @@ spec:
               name: certs
           startupProbe:
             tcpSocket:
-              port: 8080
+              port: clair-intro
             periodSeconds: 10
             failureThreshold: 300
           readinessProbe:
             tcpSocket:
               port: 8080
           livelinessProbe:
-            tcpSocket:
-              port: 8080
+            httpGet:
+              port: clair-intro
+              path: /healthz
       restartPolicy: Always
       volumes:
         - name: config


kustomize/components/clair/clair.deployment.yaml
```diff
@@ -45,8 +45,9 @@ spec:
             tcpSocket:
               port: 8080
           livelinessProbe:
-            tcpSocket:
-              port: 8080
+            httpGet:
```
alecmerdler, line 48: Can we change the `startupProbe` and `readinessProbe` to also use this port?

kustomize/components/clair/clair.deployment.yaml
```diff
@@ -45,8 +45,9 @@ spec:
...
+              port: clair-intro
```
alecmerdler, line 49: We don't currently set `introspection_addr` in Clair's config, will it still run the introspection server?

kustomize/components/clair/clair.deployment.yaml
hdonnay: I don't think that'd make sense, as the the `readinessProbe` is "ready to serve traffic", which it doesn't do on this port. The `startupProbe` could be moved over.

kustomize/components/clair/clair.deployment.yaml
hdonnay: Yes

kustomize/components/clair/clair.deployment.yaml
alecmerdler: Fixing the `startupProbe` in this way won't actually fix the problem of blocking the Quay deployment on Clair, because the `config-tool` will attempt to ping the Clair host as part of its validation step. If Clair refuses the connection (which it does currently), then the validation will fail and the Quay pod will not start.

hdonnay: Okay, well, this change encodes liveness and readiness for the pod more accurately. If `config-tool` wants to not wait for clair to be ready, that's a different change.

alecmerdler (approved):

PR #451, opened 2021-04-28
kustomize: persist DB_URI for managed postgres (PROJQUAY-1635)

alecmerdler:
Change DB_URI to be persisted rather than generated on each
reconcile. This prevents downtime on reconciles because the old and
new Quay pods continue to use the same database password.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
         run: |
           os=$(go env GOOS)
           arch=$(go env GOARCH)
-          curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+          curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_${os}_${arch}.tar.gz | tar -xz -C /tmp/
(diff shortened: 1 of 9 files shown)


kustomize/components/postgres/postgres.deployment.yaml
```diff
@@ -6,8 +6,6 @@ metadata:
     quay-component: postgres
 spec:
   replicas: 1
-  strategy:
```
alecmerdler, line 9: Note: the Postgres pod will still be recreated, but changing the strategy to `type: RollingUpdate` and persisting the password across reconciles means the database will remain available.

.github/workflows/ci.yaml
```diff
@@ -27,9 +27,8 @@ jobs:
         run: |
           os=$(go env GOOS)
           arch=$(go env GOARCH)
-          curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
-          mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
-          export PATH=$PATH:/usr/local/kubebuilder/bin
+          curl -L -o kubebuilder https://go.kubebuilder.io/dl/latest/$(go env GOOS)/$(go env GOARCH)
```
alecmerdler: Kubebuilder installation changed: https://book.kubebuilder.io/quick-start.html#installation

PR #453, opened 2021-05-05
tls: persist generated TLS cert/key pair (PROJQUAY-1838)

alecmerdler:
Move the 'ssl.cert' and 'ssl.key' to a separate, persistent
Secret to ensure that the cert/key pair is not re-generated
on every reconcile. Use k8s projected volumes to mount the
config and TLS Secrets to the same directory in the Quay
container.

### DIFF
--- a/controllers/quay/features.go
+++ b/controllers/quay/features.go
@@ -39,7 +39,11 @@ const (
 	GrafanaDashboardConfigNamespace = "openshift-config-managed"
 )
 
-func (r *QuayRegistryReconciler) checkManagedKeys(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, rawConfig []byte) (*quaycontext.QuayRegistryContext, *v1.QuayRegistry, error) {
+// FeatureDetection is a method which should return an updated `QuayRegistryContext` after performing a feature detection task.
+// TODO(alecmerdler): Refactor all "feature detection" functions to use a common function interface...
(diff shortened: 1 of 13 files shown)


kustomize/base/quay.deployment.yaml
```diff
@@ -16,9 +16,13 @@ spec:
     spec:
       serviceAccountName: quay-app
       volumes:
-        - name: configvolume
-          secret:
-            secretName: quay-config-secret
+        - name: config
```
alecmerdler, line 19: Switch to use k8s projected volume to mount both `Secrets`containing config bundle and TLS cert/key into the same `/conf/stack` directory in the container.

controllers/quay/features.go
```diff
@@ -192,9 +237,9 @@ func (r *QuayRegistryReconciler) checkObjectBucketClaimsAvailable(ctx *quayconte
 }
 
 // TODO: Improve this once `builds` is a managed component.
-func (r *QuayRegistryReconciler) checkBuildManagerAvailable(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, rawConfig []byte) (*quaycontext.QuayRegistryContext, *v1.QuayRegistry, error) {
+func (r *QuayRegistryReconciler) checkBuildManagerAvailable(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, configBundle map[string][]byte) (*quaycontext.QuayRegistryContext, *v1.QuayRegistry, error) {
```
alecmerdler, line 240: Refactored all the "feature detection" functions to accept the full config bundle so that we can check `ssl.cert`/`ssl.key`, not just `config.yaml`.

PR #454, opened 2021-05-05
api: upgrade to apiextensions.k8s.io/v1 (PROJQUAY-1791)

jonathankingfc:
- Update CRD from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1
- Migration guide: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122


alecmerdler (changes requested):
Make sure to update the CRDs in the `bundle/` directory.

alecmerdler (approved):

PR #455, opened 2021-05-05
finalizer: check permissions before finalizing (PROJQUAY-1937)

alecmerdler:
Fixes RBAC permission denied error from blocking QuayRegistry
deletion because finalizer would never be removed.


jonathankingfc (approved):
LGTM

PR #458, opened 2021-05-11
kustomize: fix unamanaged Postgres component (PROJQUAY-2002)

alecmerdler:
Fix a regression where DB_URI is not copied from provided
'configBundleSecret' when using an unmanaged database.


jonathankingfc (approved):
LGTM

PR #461, opened 2021-05-14
kustomize: remove startupProbes from Postgres pods (PROJQUAY-2010)

alecmerdler:
The startup probes seem to timeout and fail intermittently during
the lifetime of the Postgres container, even though the
container logs no problems.


jonathankingfc (approved):
LGTM

PR #462, opened 2021-05-17
add resource requests and limits

dmesser:
quay-operator: add resource requests and limits (PROJQUAY-2011)

- adds resource and requests and limits to all Quay components to help placement
- this also enables cluster auto-scaling
- prevent resource starving by setting set sane limits

### DIFF
--- a/kustomize/components/clair/clair.deployment.yaml
+++ b/kustomize/components/clair/clair.deployment.yaml
@@ -48,6 +48,13 @@ spec:
             httpGet:
               port: clair-intro
               path: /healthz
+          resources:
+            requests:
+              cpu: 2000m
+              memory: 2Gi
+            limits:
+              cpu: 4000m
+              memory: 16Gi
       restartPolicy: Always
       volumes:
         - name: config
--- a/kustomize/components/clair/postgres.deployment.yaml
+++ b/kustomize/components/clair/postgres.deployment.yaml
@@ -40,3 +40,7 @@ spec:
           volumeMounts:
             - name: postgres-data
               mountPath: /var/lib/pgsql/data
+          resources:
+            requests:
+              cpu: 500m
+              memory: 2Gi
--- a/kustomize/components/mirror/mirror.deployment.yaml
+++ b/kustomize/components/mirror/mirror.deployment.yaml
@@ -89,3 +89,10 @@ spec:
             - name: extra-ca-certs
               readOnly: true
               mountPath: /conf/stack/extra_ca_certs
+          resources:
+            requests:
+              cpu: 500m
+              memory: 512Mi
+            limits:
+              cpu: 1000m
+              memory: 2Gi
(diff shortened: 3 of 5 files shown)


kustomize/components/clair/clair.deployment.yaml
```diff
@@ -48,6 +48,13 @@ spec:
             httpGet:
               port: clair-intro
               path: /healthz
+          resources:
```
alecmerdler, line 51: How did we determine these values? Are we confident that they are sufficient?

dmesser: The minimum values are from our official docs based on past track record and represent a conservative value. Looking at my own deployments I rarely see anything below 2GiB RES memory via `top`. The limits are set to what users in the past have been considering excessive resource usage. I think the minimum we need to have here are resource requests since otherwise the percentage-based HPAs aren't going to work: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-resource-metrics

In this context we might need to rethink either the HPA percentage thresholds (need to be increased otherwise might scale out too soon) or the resource requests (need to be increased, otherwise might scale out too soon but could also complicated placement)

We could hold off the limits for now, especially memory since it would result in an OOM pod kill if the limit is reached. For scale-out resources like Quay and Clair this is probably not a problem since ideally the HPA would have scaled it out way before. For the Postgres databases this would be more detrimental but we could think about an VPA there.

alecmerdler (dismissed):

alecmerdler:
@dmesser Could you fix the commit message to conform to our standard when you get the chance?

dmesser:
@alecmerdler I thought I already fixed it but I am not sure the check was re-triggered.  In any case, do we want to proceed with requests for now and talk about limits later? requests are needed otherwhise the HPAs won't trigger.

kustomize/components/clair/clair.deployment.yaml
ricardomaraschini: @dmesser What is the status of this? Are you going to remove the limits or we go with it?

ricardomaraschini (changes requested):

kustomize/components/clair/clair.deployment.yaml
dmesser: How do you feel about keeping them except for the Postgres database(s)? All other components should be scaled out.

ricardomaraschini: Looks good to me @dmesser . Those are such high values, I believe we won't have any issue. Let's just remove the postgres limits and have it in.

dmesser:
I removed the limits for the postgres instance for now. It's a more conservative approach since we wouldn't want the database to be killed unnecessarily when the host still has resources. It would be OOM'd by the host at some point.

ricardomaraschini (approved):
/lgtm

PR #469, opened 2021-06-08
components: added tls managed component (PROJQUAY-2050)

alecmerdler:
Adds a new managed component, 'tls'. This separates the
'route' component from TLS and allows users to configure
both separately. 'EXTERNAL_TLS_TERMINATION: true' is the
opinionated setting. Managed 'tls' means the default
cluster wildcard cert is used. Unmanaged 'tls' means
the user-provided cert/key pair will be injected into
the 'Route'.


alecmerdler:
We should make sure that the [enhancement proposal](https://github.com/quay/enhancements/pull/5) is reviewed and merged before this is.

PR #478, opened 2021-06-24
Updating the name of the catalogSource image

crozzy:
I was working my way through the README and this image appears to have changed names on quay.io, it's also possible that I don't have access to something

Signed-off-by: crozzy <joseph.crosland@gmail.com>


alecmerdler:
@crozzy Thanks for catching this. The reverse is actually true: the image in the `CatalogSource` is correct but it looks like it never got pushed to quay.io. I'm going to close this, but we should fix it by building/pushing the image to quay.io.

PR #484, opened 2021-07-06
componentstatus: Reporting faulty condition for quay components (PROJQUAY-1609)

ricardomaraschini:
Reporting faulty condition for component's objects. Each component faulty conditions are reported in a separate index inside the same status property.


controllers/quay/quayregistry_status_controller.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// SetupWithManager sets up provided manager as the source of events for this reconciler.
+func (q *QuayRegistryStatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).For(&qv1.QuayRegistry{}).Complete(q)
+}
+
+// Reconcile is called for reconcile status for a given QuayRegistry components. This function
+// always rescheduled the same event on return, that makes this to run from time to time.
+func (q *QuayRegistryStatusReconciler) Reconcile(
```
jonathankingfc, line 60: Is it a common practice in operators to create a controller to report status of watched resources? Is there no way to have the QuayRegistry controller watch over its owned resources instead of just its own CR?

bundle/upstream/manifests/quayregistries.crd.yaml
```diff
@@ -53,6 +53,28 @@ spec:
           status:
             description: QuayRegistryStatus defines the observed state of QuayRegistry.
             properties:
+              componentConditions:
+                description: Component Conditions shows the conditions in a per component basis
+                type: object
+                additionalProperties:
```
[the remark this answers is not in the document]
ricardomaraschini: Yes, we could. Here the component name would be the index and as the list of components is a slice I opted to leave this as is. If we go that way then it would make sense to also change the way we set components to managed or not and use a schema there as well. WDYT?

controllers/quay/quayregistry_status_controller.go
ricardomaraschini: No, it is not a common practice (this is more of a bad practice). The QuayRegistry controller acts only when something changed in a QuayRegistry object and it does a few things that I still don't fully understand (I have been informed that it creates pods for running database upgrades, for instance). To have the QuayRegistry reconcile to happen more frequently might cause other issues and I don't feel comfortable in making such a call this close to feature freeze, hence this hacky approach to the problem. i.e. we need to review the QuayRegistry reconciliation logic before using it more frequently.

ricardomaraschini: I have been looking into this and, for example, every time the Reconcile loop of the QuayRegistry controller runs a new Secret is created (named `registry-quay-config-secret-*`) and this secret is then mounted inside a pod causing a full blown redeploy (the old secret stays there, for whatever reason it is not deleted). I imagine how many other surprises we gonna find out if we decide to run this Reconcile loop frequently.

bundle/upstream/manifests/quayregistries.crd.yaml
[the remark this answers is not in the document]
ricardomaraschini: Regardless of what we are going to do with `Spec.Components` I adjusted the status to have a schema. PTAL.

jonathankingfc (dismissed):
LGTM once commit message is fixed

BillDett (approved):

PR #486, opened 2021-07-09
build: update from downstream files (PROJQUAY-2230)

thomasmckay:
Updates from downstream files, including updated channel name


harishsurf (approved):
lgtm

PR #487, opened 2021-07-13
Refactoring Ensure and Remove OwnerReference

ricardomaraschini:
These functions do not return errors as client.Object implements
meta.Accessor all the times (otherwise we have a compile error).

By removing the returned error we can clean up the code a little bit.


ricardomaraschini:
@jonathankingfc this is not for 3.6 but for 3.6.1 instead. This can be considered an effort towards solving **PROJQUAY-2046**.

ricardomaraschini:
This has been replaced by https://github.com/quay/quay-operator/pull/496

PR #488, opened 2021-07-14
reconciler: Prevent unnecessary component enabling/disabling (PROJQUY-2198)

jonathankingfc:
- Prevent HPA, Route, and Monitoring components from being enabled/disabled based on received config.
- Block rollout on missing config fields

### DIFF
--- a/controllers/quay/quayregistry_controller_test.go
+++ b/controllers/quay/quayregistry_controller_test.go
@@ -49,6 +49,10 @@ func newConfigBundle(name, namespace string, withCerts bool) corev1.Secret {
 		"ENTERPRISE_LOGO_URL": "/static/img/quay-horizontal-color.svg",
 		"FEATURE_SUPER_USERS": true,
 		"SERVER_HOSTNAME":     "quay-app.quay-enterprise",
+		// Since the testing cluster doesn't support storage, we must mock an unmanaged storage
+		"DISTRIBUTED_STORAGE_CONFIG":            map[string]interface{}{},
+		"DISTRIBUTED_STORAGE_DEFAULT_LOCATIONS": []string{},
+		"DISTRIBUTED_STORAGE_PREFERENCE":        []string{},
 	}
 
 	data := map[string][]byte{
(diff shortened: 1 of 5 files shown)


controllers/quay/quayregistry_controller_test.go
```diff
@@ -171,31 +175,35 @@ var _ = Describe("Reconciling a QuayRegistry", func() {
 				To(Succeed())
 		})
 
-		It("should not generate a self-signed TLS cert/key pair in a new `Secret`", func() {
-			// Reconcile again to get past defaulting step
-			result, err = controller.Reconcile(context.Background(), reconcile.Request{NamespacedName: quayRegistryName})
+		// This test needs to be fixed. Since the ObjectStorage API is not available, we cannot use managed storage.
```
ricardomaraschini, line 178: Wasn't this test passing before your changes? What exactly happened that made this test to break?

pkg/kustomize/secrets_test.go
```diff
@@ -285,7 +285,7 @@ var containsComponentConfigTests = []struct {
 		"route",
 		true,
 		`SERVER_HOSTNAME: registry.skynet.com`,
-		false,
+		true,
```
ricardomaraschini, line 288: The same question is also valid here, why the behavior is different now?

controllers/quay/quayregistry_controller_test.go
jonathankingfc: This PR is to block the rollout of further resources when the provided config.yaml is invalid for the provided spec. We do not want the Operator to continue reconcliing if the given config.yaml is missing fields for unmanaged components. The testing suite in the operator is essentially running against an etcd instance, so there is no ObjectStorage API available. This means that the current tests in this step run against a spec with unmanaged object storage, and so the generated config won't have a DISTRIBUTED_STORAGE_CONFIG. This missing field will block the rollout of the TLS secret.

pkg/kustomize/secrets_test.go
jonathankingfc: This is a refactor of this function where it will only return whether or not a field for a given component is present. The logic of allowing Route fields such as `SERVER_HOSTNAME` to be set now happens outside of this function.

pkg/configure/configure.go
```diff
@@ -74,6 +74,13 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
 		// Infer managed/unmanaged components from the given `config.yaml`.
 		newComponents := []v1.Component{}
 		for _, component := range quay.Spec.Components {
+
+			// HPA and Monitoring don't have fields associated with them so we skip. Route should not change based on config either since fields are optional when managed.
+			if component.Kind == v1.ComponentHPA || component.Kind == v1.ComponentMonitoring || component.Kind == v1.ComponentRoute {
```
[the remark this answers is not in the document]
jonathankingfc: This is only in the reconfigure request coming from the config editor. You can still un/manage the component in the spec, but since these fields don't have config values associated with them, there is no way to infer whether they are managed or unmanaged based on the presence of a field in the config.

syed (approved):

PR #490, opened 2021-07-22
crdmigration: migrates crds to extensions v1 (PROJQUAY-2219)

ricardomaraschini:
v1beta1 is about to be unsupported by k8s.


crozzy (approved):
LGTM

PR #496, opened 2021-08-16
upgrade: making go routine resilient to conflicts (PROJQUAY-2395)

ricardomaraschini:
The go routine reponsible for updating the current version in status
can't fail otherwise we ended up not fully deploying Quay.

This commit makes it more resilient to transient failures. This go
routine must go away, this is a temporary fix.

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -348,7 +348,14 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
 				if upgradeJob.Status.Succeeded > 0 {
 					log.Info("Quay upgrade complete, updating `status.currentVersion`")
 
-					updatedQuay, _ := v1.EnsureRegistryEndpoint(quayContext, updatedQuay, userProvidedConfig)
+					var freshQuay v1.QuayRegistry
+					if err := r.Client.Get(ctx, req.NamespacedName, &freshQuay); err != nil {
+						log.Error(err, "could not retrieve QuayRegistry")
+						return false, err
+					}
+					qcopy := freshQuay.DeepCopy()
+
+					qcopy, _ = v1.EnsureRegistryEndpoint(quayContext, qcopy, userProvidedConfig)
 					msg := "all registry component healthchecks passing"
 					condition := v1.Condition{
 						Type:               v1.ConditionTypeAvailable,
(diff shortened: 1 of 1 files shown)


controllers/quay/quayregistry_controller.go
```diff
@@ -358,21 +365,19 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+					qcopy.Status.Conditions = v1.SetCondition(qcopy.Status.Conditions, condition)
+					qcopy.Status.CurrentVersion = v1.QuayVersionCurrent
+					r.EventRecorder.Event(qcopy, corev1.EventTypeNormal, string(v1.ConditionReasonHealthChecksPassing), msg)
 
-					if err = r.Client.Status().Update(ctx, updatedQuay); err != nil {
+					if err = r.Client.Status().Update(ctx, qcopy); err != nil {
 						log.Error(err, "could not update QuayRegistry status with current version")
-
-						return true, err
+						return false, nil
```
kleesc, line 374: Why is the error not returned anymore, and how will the controller handle that change?

controllers/quay/quayregistry_controller.go
```diff
@@ -358,21 +365,19 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 					}
 
-					updatedQuay.Spec.Components = v1.EnsureComponents(updatedQuay.Spec.Components)
-					if err = r.Client.Update(ctx, updatedQuay); err != nil {
+					qcopy.Spec.Components = v1.EnsureComponents(qcopy.Spec.Components)
+					if err = r.Client.Update(ctx, qcopy); err != nil {
 						log.Error(err, "could not update QuayRegistry spec to complete upgrade")
-
-						return true, err
+						return false, nil
```
kleesc, line 380: Ditto

controllers/quay/quayregistry_controller.go
```diff
@@ -348,7 +348,14 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
 				if upgradeJob.Status.Succeeded > 0 {
 					log.Info("Quay upgrade complete, updating `status.currentVersion`")
 
-					updatedQuay, _ := v1.EnsureRegistryEndpoint(quayContext, updatedQuay, userProvidedConfig)
+					var freshQuay v1.QuayRegistry
+					if err := r.Client.Get(ctx, req.NamespacedName, &freshQuay); err != nil {
```
kleesc, line 352: What's the reasoning for this extra request first, as opposed to using `updatedQuay` directly?

controllers/quay/quayregistry_controller.go
ricardomaraschini: The idea is to make the go routine resilient to errors by simply logging them instead of returning them. If we return an error the go routine will stop trying and this is what we are trying to avoid. As a side note, this go routine must go away because it is wrong.

controllers/quay/quayregistry_controller.go
ricardomaraschini: Same answer.

controllers/quay/quayregistry_controller.go
ricardomaraschini: If the version we have in memory is not the current version (the version in the etcd) we will receive a Conflict error. If we don't fetch a fresh version, once we receive a Conflict back, we will always see Conflicts.

kleesc (approved):
LGTM

PR #498, opened 2021-08-19
I would like to change the url's in gh-pages such that we're still able to pull the helm charts

MadsHT:


dmesser:
Keep in mind, that the Helm chart is for an outdated and unmaintained version of the Operator, that is likely not able to deploy any contemporary Quay versions anymore.

PR #499, opened 2021-08-23
database: prefer user provided database config (PROJQUAY-2415)

ricardomaraschini:
If use has provided the operator with database config, prefer to use it instead of the one created during bootstrap. Without this patch the operator was ignoring the database config provided by the user through the config tool.


crozzy (approved):
This LGTM

jonathankingfc (approved):
Approved

PR #500, opened 2021-08-24
tlscerts: keep old config bundle properties (PROJQUAY-2419)

ricardomaraschini:
Use the original configuration bundle as a starting point during the
reconfigure operation. We were losing some of the properties such as
ssl.cert and ssl.key.

With this patch we keep all config bundle properties that are not set
during the reconfigure handle execution

### DIFF
--- a/pkg/configure/configure.go
+++ b/pkg/configure/configure.go
@@ -55,21 +55,46 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
 			return
 		}
 
-		newSecret := createUpdatedSecret(reconfigureRequest)
-		if err = k8sClient.Create(context.Background(), &newSecret); err != nil {
+		var quay v1.QuayRegistry
+		nsn := types.NamespacedName{
+			Namespace: reconfigureRequest.Namespace,
+			Name:      reconfigureRequest.QuayRegistryName,
+		}
+		if err := k8sClient.Get(context.Background(), nsn, &quay); err != nil {
+			log.Error(
+				err,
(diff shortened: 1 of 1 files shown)


pkg/configure/configure.go
```diff
@@ -55,21 +55,56 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
...
+				quay.Spec.ConfigBundleSecret,
+				"namespace",
+				reconfigureRequest.Namespace,
+			)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
-		log.Info("created new config secret for QuayRegistry: " + reconfigureRequest.Namespace + "/" + reconfigureRequest.QuayRegistryName)
+		for prop, value := range oldSecret.Data {
```
crozzy: It feels to me like this logic belongs in `createUpdatedSecret`, i.e. the signature becomes: 
```go
// createUpdatedSecret takes the reconfigureRequest and the oldSecret and coalesces them into a new secret.
func createUpdatedSecret(reconfigureRequest request, oldSecret corev1.Secret) corev1.Secret
```
It would also make it easier to write tests without needing to instantiate a client etc. Thoughts?

crozzy (commented):
Just one comment, everything else LGTM

pkg/configure/configure.go
jonathankingfc: I agree

ricardomaraschini: Updated, PTAL.

crozzy (approved):
Great!

PR #501, opened 2021-08-24
database: avoid regenerating password (PROJQUAY-2319)

ricardomaraschini:
Proceed to store database root password the same way we do for the
DB_URI. By doing that we avoid postgres redeployment during every
reconcile cycle.

### DIFF
--- a/pkg/kustomize/kustomize.go
+++ b/pkg/kustomize/kustomize.go
@@ -264,6 +264,14 @@ func KustomizationFor(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistr
 		return nil, err
 	}
 
+	if ctx.DbRootPw == "" {
+		rootpw, err := generateRandomString(32)
+		if err != nil {
+			return nil, err
+		}
+		ctx.DbRootPw = rootpw
+	}
+
 	quayConfigTLSSources := []string{}
 	if ctx.ClusterWildcardCert != nil {
 		quayConfigTLSSources = append(quayConfigTLSSources, "ocp-cluster-wildcard.cert="+string(ctx.ClusterWildcardCert))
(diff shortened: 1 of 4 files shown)


pkg/kustomize/kustomize.go
```diff
@@ -324,7 +333,7 @@ func KustomizationFor(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistr
 			componentPaths = append(componentPaths, filepath.Join("..", "components", string(component.Kind)))
 			managedFieldGroups = append(managedFieldGroups, fieldGroupNameFor(component.Kind))
 
-			componentConfigFiles, err := componentConfigFilesFor(component.Kind, quay, quayConfigFiles)
+			componentConfigFiles, err := componentConfigFilesFor(ctx, component.Kind, quay, quayConfigFiles)
```
crozzy, line 336: Nit: Maybe a different name here would be a good idea, `ctx` is traditionally always `context.Context`. `qCtx` or `quayCtx` might distinguish it more.

pkg/kustomize/kustomize.go
```diff
@@ -264,6 +264,14 @@ func KustomizationFor(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistr
 		return nil, err
 	}
 
+	if ctx.DbRootPw == "" {
+		rootpw, err := generateRandomString(32)
```
crozzy, line 268: Nit: maybe `rootPw` to be consistent with the rest of the PR

crozzy (approved):
LGTM, from a code POV, just a couple nits

pkg/kustomize/kustomize.go
ricardomaraschini: Following pre established standards here. I fully agree with you.

jonathankingfc (approved):
Approved

pkg/kustomize/kustomize.go
ricardomaraschini: Yes, you are right. This was a copy and paste from a branch where I am working in refactoring this. I guess we can live with this until we land that other PR of mine.

PR #504, opened 2021-08-25
status: Only check for object bucket claim when object storage is managed (PROJQUAY-0000)

jonathankingfc:
- Only check for object bucket claim status when object storage is managed


ricardomaraschini (approved):
LGTM

PR #508, opened 2021-08-31
all: update go mod to new versions

hdonnay:
This also removes the checked-in vendor directory. This is an
anti-pattern now that modules are in wide use.


ricardomaraschini:
This won't happen in here.

PR #509, opened 2021-08-31
tlscomponent: changing TLS management state evaluation (PROJQUAY-2428)

harishsurf:
Changes the way we evaluate if TLS is managed or not. During component
check we only evalute if routes are supported by the cluster or not, but
to evalute if TLS is managed or not we inspect if user has provided
certs or not.

If user has provided TLS certs then TLS component is set as unmanaged,
it is set as managed otherwise.

(cherry picked from commit 55c03ae723187919105480f62faf166ccb8ff67b)


thomasmckay (approved):
LGTM

PR #516, opened 2021-09-08
database: avoid regenerating password (PROJQUAY-2319)

ricardomaraschini:
Proceed to store database root password the same way we do for the
DB_URI. By doing that we avoid postgres redeployment during every
reconcile cycle.

This is a cherry-pick of https://github.com/quay/quay-operator/pull/501


ricardomaraschini:
After a recent rebase by @jonathankingfc this PR is not needed anymore.

PR #517, opened 2021-09-08
tls: executing pod termination (PROJQUAY-2428)

ricardomaraschini:
If TLS is unmanaged we expect the user to provide its own cert and key.
This PR changes the route to 'passthrough' and mounts the certificates
inside quay pod if TLS is unmanaged.

### DIFF
--- a/kustomize/components/mirror/mirror.deployment.yaml
+++ b/kustomize/components/mirror/mirror.deployment.yaml
@@ -21,15 +21,15 @@ spec:
       serviceAccountName: quay-app
       volumes:
         - name: config
-          secret:
-            secretName: quay-config-secret
-        - name: extra-ca-certs
           projected:
             sources:
-              - configMap:
-                  name: cluster-service-ca
+              - secret:
+                  name: quay-config-secret
               - secret:
                   name: quay-config-tls
+        - name: extra-ca-certs
+          configMap:
+            name: cluster-service-ca
       initContainers:
         - name: quay-mirror-init
           image: quay.io/projectquay/quay@sha256:5660d7174218e1cb21bf6ef406602dbe8c01c878c630a9f310fe3e5560d4c2cd
--- a/pkg/kustomize/secrets.go
+++ b/pkg/kustomize/secrets.go
@@ -105,8 +105,10 @@ func FieldGroupFor(ctx *quaycontext.QuayRegistryContext, component v1.ComponentK
 
 		return fieldGroup, nil
 	case v1.ComponentRoute:
+		// sets tls termination in the load balancer if no cert has been provided.
+		terminateExternally := len(ctx.TLSCert) == 0 && len(ctx.TLSKey) == 0
 		fieldGroup := &hostsettings.HostSettingsFieldGroup{
-			ExternalTlsTermination: true,
+			ExternalTlsTermination: terminateExternally,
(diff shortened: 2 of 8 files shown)


pkg/middleware/middleware.go
```diff
@@ -40,19 +41,33 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+	// if we are not managing TLS then user has provided its own key pair. on this case we
+	// set up the route as passthrough thus delegating TLS termination to the pods. as quay
+	// builders route uses GRPC we do not change its route's target port.
+	rt.Spec.TLS = &route.TLSConfig{
+		Termination:                   route.TLSTerminationPassthrough,
+		InsecureEdgeTerminationPolicy: route.InsecureEdgeTerminationPolicyRedirect,
+	}
+	if quayComponentLabel == "quay-app-route" {
+		rt.Spec.Port = &route.RoutePort{
+			TargetPort: intstr.Parse("https"),
```
crozzy, line 67: I'm not sure I 100% understand this, `intstr.Parse()` is going to create an object whose `String()` method will return `"https"` and whose `IntValue()` method will return `0`?

crozzy (commented):
LGTM, just one question

pkg/middleware/middleware.go
ricardomaraschini: Yeah, it is a struct (`intstr.IntOrString`) used by k8s when a field may be an int or a string (like the port).

pkg/kustomize/secrets.go
```diff
@@ -105,8 +105,10 @@ func FieldGroupFor(ctx *quaycontext.QuayRegistryContext, component v1.ComponentK
 
 		return fieldGroup, nil
 	case v1.ComponentRoute:
+		// sets tls termination in the load balancer if no cert has been provided.
+		terminateExternally := len(ctx.TLSCert) == 0 && len(ctx.TLSKey) == 0
 		fieldGroup := &hostsettings.HostSettingsFieldGroup{
-			ExternalTlsTermination: true,
+			ExternalTlsTermination: terminateExternally,
 			PreferredUrlScheme:     "https",
```
jonathankingfc, line 112: Should this also be switched on whether or not we are terminating externally?

kustomize/components/mirror/mirror.deployment.yaml
```diff
@@ -21,15 +21,15 @@ spec:
...
             sources:
-              - configMap:
-                  name: cluster-service-ca
+              - secret:
+                  name: quay-config-secret
               - secret:
                   name: quay-config-tls
+        - name: extra-ca-certs
+          configMap:
+            name: registry-cluster-service-ca
```
ricardomaraschini: this is wrong.

pkg/kustomize/secrets.go
ricardomaraschini: As we checked, it seems like we don't need it to be switched to http.

kustomize/components/mirror/mirror.deployment.yaml
ricardomaraschini: now this is right.

jonathankingfc (approved):
LGTM

PR #518, opened 2021-09-09
tls: Check for certs to mark tls as unmanaged (PROJQUAY-2428)

jonathankingfc:
- TLS should determined to be unmanaged by the presence of tls cert and key, not on any particular field
- Fields from the config.yaml are no longer used to determine this

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -244,10 +244,18 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
+	userProvidedCerts := map[string][]byte{}
+	if _, ok := configBundle.Data["ssl.cert"]; ok {
+		userProvidedCerts["ssl.cert"] = configBundle.Data["ssl.cert"]
+	}
+	if _, ok := configBundle.Data["ssl.key"]; ok {
+		userProvidedCerts["ssl.key"] = configBundle.Data["ssl.key"]
+	}
+
 	updatedQuay.Status.Conditions = v1.RemoveCondition(updatedQuay.Status.Conditions, v1.ConditionTypeRolloutBlocked)
 
 	for _, component := range updatedQuay.Spec.Components {
-		contains, err := kustomize.ContainsComponentConfig(userProvidedConfig, component)
+		contains, err := kustomize.ContainsComponentConfig(userProvidedConfig, userProvidedCerts, component)
 		if err != nil {
 			updatedQuay, err = r.updateWithCondition(&quay, v1.ConditionTypeRolloutBlocked, metav1.ConditionTrue, v1.ConditionReasonConfigInvalid, err.Error())
 			if err != nil {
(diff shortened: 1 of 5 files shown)


controllers/quay/quayregistry_controller.go
```diff
@@ -244,10 +244,36 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			updatedQuay, err = r.updateWithCondition(&quay, v1.ConditionTypeRolloutBlocked, metav1.ConditionTrue, v1.ConditionReasonConfigInvalid, err.Error())
+			if err != nil {
+				log.Error(err, "failed to update `conditions` of `QuayRegistry`")
+
+				return ctrl.Result{}, nil
+			}
+		}
+	}
+	userProvidedCerts := map[string][]byte{"ssl.key": userProvidedTLSKey, "ssl.cert": userProvidedTLSCert}
+
```
ricardomaraschini, line 254: This block seems useless. Secret data field is already a `map[string][]byte` and there is no need to unmarshal its content. What is the reasoning behind it?

controllers/quay/quayregistry_controller.go
```diff
@@ -244,10 +244,36 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	updatedQuay.Status.Conditions = v1.RemoveCondition(updatedQuay.Status.Conditions, v1.ConditionTypeRolloutBlocked)
 
 	for _, component := range updatedQuay.Spec.Components {
-		contains, err := kustomize.ContainsComponentConfig(userProvidedConfig, component)
+		contains, err := kustomize.ContainsComponentConfig(userProvidedConfig, userProvidedCerts, component)
```
ricardomaraschini, line 258: We could pass in `configBundle.Data` directly, no?

ricardomaraschini (changes requested):

controllers/quay/quayregistry_controller.go
jonathankingfc: Resolved

controllers/quay/quayregistry_controller.go
jonathankingfc: Not unless we alter how it is called elsewhere as well, since the call to this function in configure.go takes a different shape as configBundle.Data.

ricardomaraschini: I don't follow, `configBundle.Data` is of type `map[string][]byte`. `userProvidedCerts` is also a `map[string][]byte`.

jonathankingfc: We can do this, but we will then have to combine reconfigureRequest.Certs and reconfigureRequest.Config into a single map in configure.go

ricardomaraschini: I don't follow, sorry. This makes no sense to me. The code "copies" data from A (`configBundle.Data`) to B (`userProvidedCerts`)  if data is present in A, at last it then passes B to the function instead of just passing A and avoid the "copy" all together. Why can't we pass A directly here? As far as I understand nothing needs to be changed in configure.go.

ricardomaraschini (changes requested):

ricardomaraschini (approved):

PR #519, opened 2021-09-09
upgrade: Upgrade rbac version to v1 (PROJQUAY-2516)

jonathankingfc:
- Kubernetes v1.22 no longer supports v1beta1 RBAC kind, so it must be upgraded in order to support Openshift 4.9.
- See deprecation notice here -> https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122


harishsurf (approved):
lgtm

PR #520, opened 2021-09-09
hpa: Only deploy clair HPA when HPA component is managed (PROJQUAY-2435)

jonathankingfc:
- Clair HPA is only deployed when HPA component is managed


ricardomaraschini (changes requested):
You gonna need take this change into account when reporting the HPA status condition [here](https://github.com/quay/quay-operator/blob/master/pkg/cmpstatus/hpa.go). Remember to also add a test case for it. You may use [this](https://github.com/quay/quay-operator/blob/master/pkg/cmpstatus/clair.go#L45-L92) as a base as it also checks for multiple objects of the same kind (deployments in this case).

ricardomaraschini (approved):
lgtm

PR #524, opened 2021-09-16
fix(bundle): use correct channel and operator name in subscription (PROJQUAY-2556)

flavianmissi:
the previously used values did not match the ones specified in the upstream manifests.

### DIFF
--- a/bundle/quay-operator.subscription.yaml
+++ b/bundle/quay-operator.subscription.yaml
@@ -3,7 +3,7 @@ kind: Subscription
 metadata:
   name: quay-operator-tng
 spec:
-  channel: preview-3.6
+  channel: alpha
   installPlanApproval: Automatic
   name: quay-operator-tng
   source: quay-operator-tng
--- a/bundle/quay-operator.subscription.yaml
+++ b/bundle/quay-operator.subscription.yaml
@@ -1,11 +1,11 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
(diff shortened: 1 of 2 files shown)


bundle/quay-operator.subscription.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
-  name: quay-tng
+  name: quay-operator-tng
 spec:
-  channel: alpha
+  channel: preview-3.6
```
jonathankingfc: I would leave this version agnostic unless we want to update this with each release. Could we somehow pull this in from the CSV as an annotation? Otherwise LGTM

flavianmissi: Good point! I changed this back to alpha and changed the upstream annotations to match.

flavianmissi:
@jonathankingfc do you have time to give this another review round?

crozzy (approved):
LGTM

crozzy:
@flavianmissi seems like this repo is like `quay/quay` where you need a jira issue in the commit message

flavianmissi:
> @flavianmissi seems like this repo is like `quay/quay` where you need a jira issue in the commit message

@crozzy how do go about it when there is no jira issue? create one for the sake of the commit check passing?

BillDett:
@flavianmissi - you can reference one of the Jiras you're using already or just make a new one.

PR #525, opened 2021-09-16
tls: executing pod termination (PROJQUAY-2428)

ricardomaraschini:
If TLS is unmanaged we expect the user to provide its own cert and key.
This PR changes the route to 'passthrough' and mounts the certificates
inside quay pod if TLS is unmanaged.


jonathankingfc (approved):
LGTM

PR #530, opened 2021-09-20
chore: github workflow to publish container images (PROJQUAY-2556)

flavianmissi:
For now this targets branch `redhat-3.5`, but we can reuse it for any branch (as long as we update the image tags accordingly). The end goal is to use the builds produced by this workflow on a nightly job that will deploy Quay using this operator.

One important thing to notice is that related projects (clair, quay-builder, etc) also need to publish an equivalent `3.5` image. I'll have PRs open and merged for these projects before merging this one. 

Example of a [successful workflow](https://github.com/flavianmissi/quay-operator/actions/runs/1257257820) (using my own registry instead of `projectquay`).

### DIFF
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -1,0 +1,169 @@
+---
+name: Build and Publish Images
+
+on:
+  push:
+    branches:
+      - redhat-3.5
+
+jobs:
+  quay-image:
+    name: Calculate Quay Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      IMAGE_REGISTRY: quay.io/projectquay
+      TAG: 3.5-unstable
+    steps:
+      - name: Pull Image
+        id: pull-image
+        run: docker pull "${IMAGE_REGISTRY}"/quay:"${TAG}"
+      - name: Set Output
+        id: set-output
+        run: echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE_REGISTRY}/quay:${TAG})"
+
+  clair-image:
+    name: Calculate Clair Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      IMAGE_REGISTRY: quay.io/projectquay
+      TAG: 3.5-unstable
+    steps:
+      - name: Pull Image
+        id: pull-image
+        run: docker pull "${IMAGE_REGISTRY}"/clair:"${TAG}"
+      - name: Set Output
+        id: set-output
+        run: echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE_REGISTRY}/clair:${TAG})"
+
+  builder-image:
+    name: Calculate Builder Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      IMAGE_REGISTRY: quay.io/projectquay
+      TAG: 3.5-unstable
+    steps:
+      - name: Pull Image
+        id: pull-image
+        run: docker pull "${IMAGE_REGISTRY}"/quay-builder:"${TAG}"
+      - name: Set Output
+        id: set-output
+        run: echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE_REGISTRY}/quay-builder:${TAG})"
+
+  qemu-builder-image:
+    name: Calculate Qemu Builder Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      IMAGE_REGISTRY: quay.io/projectquay
+    steps:
+      - name: Pull Image
+        id: pull-image
+        run: docker pull "${IMAGE_REGISTRY}"/quay-builder-qemu:latest
+      - name: Set Output
+        id: set-output
+        run: echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE_REGISTRY}/quay-builder-qemu:latest)"
+
+  operator-image:
+    name: Publish Operator Image
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      OPERATOR_IMAGE: quay.io/projectquay/quay-operator:3.5-unstable
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: ${{ env.OPERATOR_IMAGE }}
+
+      - name: Set digest output
+        id: set-output
+        run: |
+          docker pull "${OPERATOR_IMAGE}"
+          echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${OPERATOR_IMAGE})"
+
+  operator-index-images:
+    name: Publish Catalog Index Image
+    runs-on: 'ubuntu-latest'
+    needs: [quay-image, clair-image, builder-image, qemu-builder-image, operator-image]
+    env:
+      OPERATOR_NAME: quay-operator-test
+      BUNDLE: quay.io/projectquay/quay-operator-bundle
+      INDEX: quay.io/projectquay/quay-operator-index
+      TAG: 3.5-unstable
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Update CSV Image Versions
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq eval -i '
+              .metadata.name = strenv(OPERATOR_NAME) |
+              .metadata.annotations.quay-version = strenv(TAG) |
+              .metadata.annotations.containerImage = "${{needs.operator-image.outputs.digest}}" |
+              del(.spec.replaces) |
+              .spec.install.spec.deployments[0].name = strenv(OPERATOR_NAME) |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].image = "${{needs.operator-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[4].value = "${{needs.quay-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[5].value = "${{needs.clair-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[6].value = "${{needs.builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[7].value = "${{needs.qemu-builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[8].value = "centos/postgresql-10-centos7@sha256:de1560cb35e5ec643e7b3a772ebaac8e3a7a2a8e8271d9e91ff023539b4dfb33" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[9].value = "centos/redis-32-centos7@sha256:06dbb609484330ec6be6090109f1fa16e936afcf975d1cbc5fff3e6c7cae7542"
+              ' bundle/downstream/manifests/quay-operator.clusterserviceversion.yaml
+
+      - name: Update Bundle Annotations
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq eval -i '
+              .annotations."operators.operatorframework.io.bundle.channel.default.v1" = "test" |
+              .annotations."operators.operatorframework.io.bundle.channels.v1" = "test" |
+              .annotations."operators.operatorframework.io.bundle.package.v1" = strenv(OPERATOR_NAME)
+              ' bundle/downstream/metadata/annotations.yaml
+
+      - name: Publish Bundle Image
+        uses: docker/build-push-action@v2
+        with:
+          context: ./bundle/downstream
+          file: ./bundle/Dockerfile
+          push: true
+          tags: ${{ env.BUNDLE }}:${{ env.TAG }}
+
+      - name: Publish Catalog Index
+        env:
+          OPM_DOWNLOAD_URL: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.6/
+          OPM_TAR: opm-linux.tar.gz
+        run: |
+          wget -q "${OPM_DOWNLOAD_URL}/${OPM_TAR}"
+          tar xvf "${OPM_TAR}"
+          ./opm index add --build-tool docker --bundles "${BUNDLE}:${TAG}" --tag "${INDEX}:${TAG}"
+          docker push "${INDEX}:${TAG}"


.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+            yq eval -i '
+              .metadata.annotations.quay-version = "env(Y_STREAM)" |
+              .metadata.annotations.containerImage = "quay.io/projectquay/quay-operator:env(Y_STREAM)" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].image = "quay.io/projectquay/quay-operator:env(Y_STREAM)" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[4].value = "${{needs.quay-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[5].value = "${{needs.clair-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[6].value = "${{needs.builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[7].value = "${{needs.qemu-builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[8].value = "${{needs.postgres-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[9].value = "${{needs.redis-image.outputs.digest}}"
```
flavianmissi: These lines are fragile (`env[4-9]`). Using annotations (i.e. `.metadata.annotations.quay-version = "env(Y_STREAM)"`) is a better idea, but since it requires changes to the CSV yaml on both `master` and `redhat-3.5` I'll do that in a separate PR.

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+    runs-on: 'ubuntu-latest'
+    env:
+      OPERATOR_TAG: quay.io/projectquay/quay-operator:3.5
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          ref: 'redhat-3.5'
+
+      - name: Login to Quay.io
```
ricardomaraschini, line 84: Can we use `docker/login-action@v1` here instead?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+          QUAY_USER: projectquay+quay_github
+        run: |
+          echo "::add-mask::${{ secrets.QUAY_TOKEN }}"
+          docker login -u "${QUAY_USER}" -p '${{ secrets.QUAY_TOKEN }}' quay.io
+
+      - name: Build Operator
+        run: docker build -t "${OPERATOR_TAG}" .
+
+      - name: Push Operator Image
+        run: docker push "${OPERATOR_TAG}"
```
ricardomaraschini: Can we use `docker/build-push-action@v2` for building and pushing instead?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+    env:
+      INDEX_TAG: quay.io/projectquay/quay-operator-index:3.5
+      Y_STREAM: 3.5
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          ref: 'redhat-3.5'
+
+      - name: Login to Quay.io
```
ricardomaraschini, line 117: dito

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+  postgres-image:
+    name: Calculate Postgres Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    steps:
+      - name: Set Output
+        id: set-output
+        run: echo "::set-output name=digest::centos/postgresql-10-centos7@sha256:de1560cb35e5ec643e7b3a772ebaac8e3a7a2a8e8271d9e91ff023539b4dfb33"
```
ricardomaraschini: Are we doing this to keep the steps similar? This is going to always have the same "output", no?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,157 @@
...
+          echo "::add-mask::${{ secrets.QUAY_TOKEN }}"
+          docker login -u "${QUAY_USER}" -p '${{ secrets.QUAY_TOKEN }}' quay.io
+
+      - name: Update CSV Image Versions
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq eval -i '
+              .metadata.annotations.quay-version = strenv(Y_STREAM) |
+              .metadata.annotations.containerImage = "quay.io/fmissi/quay-operator:"+env(Y_STREAM) |
```
ricardomaraschini: Missing image update here?

.github/workflows/build-and-publish.yaml
flavianmissi: I didn't know that action! Will fix 👍

.github/workflows/build-and-publish.yaml
flavianmissi: Yes and yes. I leaned on homogeneity here... but now that you point it out it feels overkill 🤔

.github/workflows/build-and-publish.yaml
flavianmissi: Good catch!

.github/workflows/build-and-publish.yaml
ricardomaraschini: Assuming that we would like to use the "last 3.5 build" of quay, clair and builder why not replace only the image tag to point to `v3.5`? i.e. we would not need to know what is the current sha. wdyt?

.github/workflows/build-and-publish.yaml
flavianmissi: done (result: https://github.com/flavianmissi/quay-operator/actions/runs/1257661839)

.github/workflows/build-and-publish.yaml
flavianmissi: moved the image digests for redis and postgres to where they are used.

.github/workflows/build-and-publish.yaml
flavianmissi: That's what I tried first, I moved to using the sha because the operator refused to deploy quay via tag only. Not sure if this could behavior be disabled though.

ricardomaraschini: Yes, you are right: https://github.com/quay/quay-operator/blob/redhat-3.5/pkg/kustomize/kustomize.go#L73-L75

flavianmissi: Ah thanks for finding the source!

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+jobs:
+  quay-image:
+    name: Calculate Quay Image Digest
+    runs-on: 'ubuntu-latest'
+    outputs:
+      digest: ${{ steps.set-output.outputs.digest }}
+    env:
+      IMAGE_REGISTRY: quay.io/projectquay
+      Y_STREAM: 3.5
```
jonathankingfc: should we just make this TAG? that way we can include things like latest, nightly, etc.

flavianmissi: good idea!

flavianmissi: As for the value itself, do we want `3.5`, `v3.5`, or even `redhat-3.5`?
To me the only important thing is that we're consistent throughout repositories. If any project uses different versions for upstream vs downstream then we might want to go with `redhat-3.5`, but I don't know if this is the case (maybe clair does this?)

flavianmissi: Using the branch name (`redhat-3.5`) also has the benefit of it being easy to automate for different branches, I think github actions provides means of retrieving that via environment variable.

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          ref: 'redhat-3.5'
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: projectquay+quay_github
```
harishsurf: nit: would it be a good idea to have this coming from an `env` variable as well instead of being hardcoded?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,136 @@
...
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Update CSV Image Versions
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq eval -i '
+              .metadata.annotations.quay-version = strenv(TAG) |
```
harishsurf, line 130: we should also edit the following fields: 
`.metadata.name`
`.spec.version`
`.spec.install.spec.deployments.name`

Note: The CSV also has a `replaces` field which specifies the previous operator version to be replaced ([example](https://dist-git.host.prod.eng.bos.redhat.com/cgit/containers/quay-operator-bundle/tree/manifests/quay-operator.clusterserviceversion.yaml?h=quay-3.5-rhel-8#n319)). This only applies to a z-stream. For a y-stream, this field is commented out. ([related doc](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#if-using-replaces))

If the end goal of this workflow is to just use the index image to test out quay-operator in a **clean** cluster, then we could skip the `replaces` field completely (making it a y-stream only)

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,136 @@
...
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[4].value = "${{needs.quay-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[5].value = "${{needs.clair-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[6].value = "${{needs.builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[7].value = "${{needs.qemu-builder-image.outputs.digest}}" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[8].value = "$centos/postgresql-10-centos7@sha256:de1560cb35e5ec643e7b3a772ebaac8e3a7a2a8e8271d9e91ff023539b4dfb33" |
+              .spec.install.spec.deployments[0].spec.template.spec.containers[0].env[9].value = "centos/redis-32-centos7@sha256:06dbb609484330ec6be6090109f1fa16e936afcf975d1cbc5fff3e6c7cae7542"
+              ' deploy/manifests/quay-operator/"${TAG}"/quay-operator.clusterserviceversion.yaml
+
+      - name: Build Catalog Index
```
harishsurf: The image built in this step is the `Catalog Bundle Image` or `bundle image`. In order to create the `index image`, we would have to use the `opm index add --bundles` command. The [readme doc](https://github.com/quay/quay-operator) (from step 4 - till end) has the steps.

harishsurf (changes requested):
looks really good! left some comments. Let me know if you have questions

.github/workflows/build-and-publish.yaml
flavianmissi: Thank you, I'll replace the fields I missed. We can definitely drop the `replaces`!

.github/workflows/build-and-publish.yaml
flavianmissi: As far as I understand that's only the case for the new bundle model we're using from 3.6. I followed the [README for 3.5](https://github.com/quay/quay-operator/tree/redhat-3.5#local-development) (see `Building custom CatalogSource` section) to build the package manifest under `deploy`, which seems to entirely skip using `opm`. Did I miss something?

.github/workflows/build-and-publish.yaml
flavianmissi: Indeed this should probably be a secret, would be better for test running the workflow too.

flavianmissi: Done. Also configured the actual secret in the repo so we're good to go.

.github/workflows/build-and-publish.yaml
flavianmissi: all done!

.github/workflows/build-and-publish.yaml
flavianmissi: had to change this, turns out the CSV on `redhat-3.5` isn't working... I suppose we don't we it for anything? 
marking this as resolved as I'm doing as suggested and using opm index.

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Update CSV Image Versions
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq eval -i '
+              .metadata.name = "quay-operator-test" |
+              .metadata.annotations.quay-version = strenv(TAG) |
+              .metadata.annotations.containerImage = "quay.io/projectquay/quay-operator:"+env(TAG) |
+              .spec.version = strenv(TAG)+".999" |
```
flavianmissi: Note: building with `opm` requires a semver, to make it happy I added an unlike patch version here. lmk if there's a better way.

.github/workflows/build-and-publish.yaml
harishsurf: I am not aware how the operator was deployed prior to `3.6`. I was able to use `opm` cli to create the index image for a 3.5.8 operator(using the CSV for redhat-3.5). Let's follow up the conversation on [slack](https://coreos.slack.com/archives/CGJC7J7KR/p1632411594250300)

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,159 @@
...
+    runs-on: 'ubuntu-latest'
+    needs: [quay-image, clair-image, builder-image, qemu-builder-image]
+    env:
+      BUNDLE: quay.io/projectquay/quay-operator-bundle
+      INDEX: quay.io/projectquay/quay-operator-index
+      TAG: 3.5
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        # runs on main branch to use newer OLM bundle files.
```
ricardomaraschini: I guess this will checkout the branch where the push happened (on this case `redhat-3.5`), no?

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,159 @@
...
+        run: echo "::set-output name=digest::$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE_REGISTRY}/quay-builder-qemu:latest)"
+
+  operator-image:
+    name: Publish 3.5 Operator Image
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          ref: 'redhat-3.5' # must use correct version branch!
```
ricardomaraschini: This seems redundant with 
```
on:
  push:
    branches:
      - redhat-3.5
```
Not a deal breaker but seems odd we have to enforce this.

.github/workflows/build-and-publish.yaml
```diff
@@ -0,0 +1,159 @@
...
+            yq eval -i '
+              .annotations."operators.operatorframework.io.bundle.channel.default.v1" = "test" |
+              .annotations."operators.operatorframework.io.bundle.channels.v1" = "test" |
+              .annotations."operators.operatorframework.io.bundle.package.v1" = "quay-operator-test"
+              ' bundle/downstream/metadata/annotations.yaml
+
+      - name: Publish Bundle Image
+        run: |
+          docker build -t "${BUNDLE}:${TAG}" -f ./bundle/Dockerfile ./bundle/downstream
+          docker push "${BUNDLE}:${TAG}"
```
ricardomaraschini: docker/build-push-action@v2

flavianmissi: I looked into using that action here too, but the docs states:

> Be careful because any file mutation in the steps that precede the build step will be ignored [...]

Thus we can't use it, since we want the changes on the previous step to be present in the image we're building.

flavianmissi: https://github.com/docker/build-push-action#usage

ricardomaraschini: Oh, interesting. I was not aware of that. I guess we could still use `context: .` though. Anyways, I am OK with whatever you decide.

flavianmissi: Ha! I guess I didn't finished reading the paragraph, nice catch. I'll try the context.

harishsurf (dismissed):
lgtm!

.github/workflows/build-and-publish.yaml
flavianmissi: Talked about this IRL - resolving the conversation

.github/workflows/build-and-publish.yaml
flavianmissi: Talked about this IRL - resolving the conversation

.github/workflows/build-and-publish.yaml
flavianmissi: tried this and couldn't get working on first try - I'll leave it as is for now just so I can move on.

.github/workflows/build-and-publish.yaml
flavianmissi: I just figured this out with Ricardo's help @harishsurf. It works for you because we're using the bundle format for downstream. https://github.com/quay/quay-operator/pull/538 makes the same (or similar enough anyway) changes to `redhat-3.5` so that we can use the `opm` to build upstream too :)

flavianmissi:
I have changed a few things (and retested it all), please have another look!

ricardomaraschini (dismissed):

dmage (approved):

PR #531, opened 2021-09-22
mirror: Set mirror as managed when flag enabled in editor (PROJQUAY-2489)

jonathankingfc:
- Enabled mirror component when repository mirroring flag is enabled through config editor

### DIFF
--- a/pkg/configure/configure.go
+++ b/pkg/configure/configure.go
@@ -100,18 +100,27 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
 		newComponents := []v1.Component{}
 		for _, component := range quay.Spec.Components {
 
+			var contains bool
 			// HPA and Monitoring don't have fields associated with them so we skip. Route should not change based on config either since fields are optional when managed.
 			if component.Kind == v1.ComponentHPA || component.Kind == v1.ComponentMonitoring || component.Kind == v1.ComponentRoute {
 				newComponents = append(newComponents, component)
 				continue
 			}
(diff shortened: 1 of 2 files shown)


pkg/configure/configure.go
```diff
@@ -100,20 +100,29 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
...
 			if err != nil {
 				log.Error(err, "failed to check `config.yaml` for component fieldgroup", "component", component.Kind)
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
 
+			// Mirror we infer based on the value of FEATURE_REPO_MIRROR
+			if component.Kind == v1.ComponentMirror {
+				enabled, ok := reconfigureRequest.Config["FEATURE_REPO_MIRROR"]
+				if ok && enabled.(bool) {
```
ricardomaraschini, line 121: This will panic if FEATURE_REPO_MIRROR is anything other than a boolean, no?

ricardomaraschini (changes requested):

pkg/configure/configure.go
jonathankingfc: This will only be hit after a config validation, so we can assume that the config schema is correct already.

ricardomaraschini: Where is this config validation made? As far as my understanding of this code goes it is just a `map[string]interface{}` at this stage.

ricardomaraschini (approved):

PR #532, opened 2021-09-22
[WIP] webhook: Add validating and mutating webhooks

jonathankingfc:
- Add mutating webhook to ensure defaults on the CRD
- Add validating webhook to validate spec in with provided config bundle


ricardomaraschini:
@jonathankingfc shall we close this one?

PR #535, opened 2021-09-23
feat: setup E2E nightly run (PROJQUAY-2556)

flavianmissi:
Successful run: https://github.com/flavianmissi/quay-operator/runs/3688507443?check_suite_focus=true

This has a few quirks that hopefully using something like [kuttl](https://kuttl.dev) will help us with, for instance the workflow cannot be run in parallel, as the namespace isn't created during build. I'm thinking we can live with this limitation for now, since the workflow will run on a schedule.

There are a few TODOs before merging, but they don't block review. Namely:
 * [x] replace currently used namespace with a better named one (has to be created in cluster first)
 * [x] replace my own quay.io username with `projectquay`

Note that this has to be merged after https://github.com/quay/quay-operator/pull/530

### DIFF
--- a/.github/workflows/e2e-nightly.yaml
+++ b/.github/workflows/e2e-nightly.yaml
@@ -1,0 +1,166 @@
+---
+name: E2E Nightly
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to run tests from. Defaults to redhat-3.5'
+        required: false
+      tag:
+        description: 'quay-operator-index tag. Defaults to 3.5-unstable'
+        required: false
+  schedule:
+    - cron: '30 5 * * *'
+
+jobs:
+  deploy:
+    name: Deploy the operator
+    runs-on: 'ubuntu-latest'
+    env:
+      BRANCH: ${{ github.event.inputs.branch || 'redhat-3.5' }}
+      TAG: ${{ github.event.inputs.tag || '3.5-unstable' }}
+      CATALOG_PATH: ./bundle/quay-operator.catalogsource.yaml
+      OG_PATH: ./bundle/quay-operator.operatorgroup.yaml
+      SUBSCRIPTION_PATH: ./bundle/quay-operator.subscription.yaml
+      QUAY_SAMPLE_PATH: ./config/samples/managed.quayregistry.yaml
+      OPERATOR_PKG_NAME: quay-operator-test
+      NAMESPACE: quay-operator-e2e-nightly
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.BRANCH }}
+
+      - name: Setup catalog source yaml
+        uses: mikefarah/yq@master
+        env:
+          CATALOG_IMAGE: quay.io/projectquay/quay-operator-index:${{ env.TAG }}
+        with:
+          cmd: |
+            yq e -i '
+              .spec.image = env(CATALOG_IMAGE) |
+              .metadata.name = env(OPERATOR_PKG_NAME)
+              ' ${CATALOG_PATH}
+
+      - name: Create catalog source
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.KUBE_CONFIG }}
+        with:
+          args: create -n openshift-marketplace -f ${{ env.CATALOG_PATH }}
+
+      - name: Setup operator group yaml
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq e -i '.spec.targetNamespaces[0] = env(NAMESPACE)' ${OG_PATH}
+
+      - name: Create operator group
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.KUBE_CONFIG }}
+        with:
+          args: create -n ${{ env.NAMESPACE }} -f ${{ env.OG_PATH }}
+
+      - name: Setup subscription yaml
+        uses: mikefarah/yq@master
+        with:
+          cmd: |
+            yq e -i '
+              .spec.channel = "test" |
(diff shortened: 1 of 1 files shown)


jonathankingfc (dismissed):
LGTM, just make sure to squash your commits into one.

.github/workflows/e2e-nightly.yaml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Setup operator group yaml
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq e -i '.spec.targetNamespaces[0] = env(NAMESPACE)' ${OG_PATH}
+
+      - name: Create operator group
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.KUBE_CONFIG }}
```
harishsurf, line 61: Fyi: Alternative to using OCP cluster, there is [kind](https://kind.sigs.k8s.io/) tool (came across this during olm time ([example](https://github.com/operator-framework/operator-lifecycle-manager/blob/2fd014019ca8cc86ae859da797e814b72c299257/.github/workflows/e2e-kind.yml#L16)) which lets you spin up cluster in <2min. We probably need to:
- install olm using the `operator-sdk` cli ([doc](https://sdk.operatorframework.io/docs/cli/operator-sdk_olm_install/))
- create `crds` on the cluster before hand
I see gatekeeper operator also does something similar [here](https://github.com/gatekeeper/gatekeeper-operator/blob/master/.github/workflows/olm_tests.yaml#L47-L65)
This reduces overhead on the OCP cluster and could get away with not having to cleanup resources

Also, noob me: Does quay even work on a non-openshift cluster?

harishsurf (commented):
lgtm!

.github/workflows/e2e-nightly.yaml
flavianmissi: Yeah I heard of kind! This being an end-to-end test I think there is a point in running it against a proper openshift cluster. If/when we end up in openshift CI, then we will have the CI itself bootstrapping clusters for us. In the meanwhile we can use [kuttl](https://kuttl.dev) (credit to Jonathan for presenting it to the team) which I think could help us getting stuff setup/cleanup more easily.

And I'm a noob here too, but can't see why both quay and the operator wouldn't work in any k8s cluster as long as olm is installed(:

BillDett: Quay itself works fine on k8s, but with the Operator we need OpenShift as we rely on Route objects (which are OCP specific) for ingress.

flavianmissi: Oh! Does it work on [OKD](https://www.okd.io/) then?

dmage: It should, OKD is a community supported version of OpenShift.

.github/workflows/e2e-nightly.yaml
```diff
@@ -0,0 +1,156 @@
...
+        with:
+          args: rollout -n ${{ env.NAMESPACE }} status deployment skynet-quay-app --timeout=3m
+
+      - name: Ensure mirror rollout
+        uses: actions-hub/kubectl@master
+        env:
+          KUBE_CONFIG: ${{ secrets.KUBE_CONFIG }}
+        with:
+          args: rollout -n ${{ env.NAMESPACE }} status deployment skynet-quay-mirror --timeout=3m
```
dmage, line 126: Can you add a step that pushes/pulls something from Quay? Or some kind of stub for actual e2e tests.

.github/workflows/e2e-nightly.yaml
BillDett: I think you'll find it safer to stick with OCP itself, not sure how up to date OKD stays with OCP nightlies...

flavianmissi: Definitely @BillDett , I'm asking mostly out of curiosity ([and because our docs seem to be misleading about this](https://github.com/quay/quay-operator#getting-started), perhaps it has worked previously 🤔

.github/workflows/e2e-nightly.yaml
flavianmissi: I think this would be a nice addition @dmage! I'm not sure exactly which direction to go here however. Since these are e2e tests for the operator itself we'd want the test cases to focus on its own functionality. 
Depending on how good Quay's health check is we might not have to perform any operations against it to know it works.

There's also [`kuttl`](https://kuttl.dev/) that @jonathankingfc brought up, which we can use to build better coverage as we iterate upon our suite.

jonathankingfc (dismissed):
LGTM

.github/workflows/e2e-nightly.yaml
dmage: Health checks is a good option, we'll need to check them in the Quay tests so that we can rely on them. Perhaps at this point we can simply rely on the operator's reportings.

I'd like to see there a step that will execute a shell script/makefile target from the repository. At this point, it can be a simple check that the deployment is available. But it'll give us a check that every engineer can run on their cluster.

flavianmissi: > Health checks is a good option, we'll need to check them in the Quay tests so that we can rely on them. Perhaps at this point we can simply rely on the operator's reportings.

Agreed. I'll have a look at this and ping you with my findings.

> I'd like to see there a step that will execute a shell script/makefile target from the repository. At this point, it can be a simple check that the deployment is available. But it'll give us a check that every engineer can run on their cluster.

I really like the idea of making commands used in the e2e suite available for anyone to run against their own cluster. From what I gather we can use kuttl for that too - it requires a [kubectl plugin](https://kuttl.dev/docs/#install-kuttl-cli), but I think that isn't so much to ask from a dev wanting to run the e2e suite.
There's a [JIRA](https://issues.redhat.com/browse/PROJQUAY-2594) tracking the work on adding kuttl to the suite.

dmage (approved):

PR #536, opened 2021-09-23
redis: Mark Redis as a required component (PROJQUAY-2455)

jonathankingfc:
- Set Redis as a required component for a QuayRegistry deployment
- Block reconcile when config.yaml is missing fields for unmanaged Redis


harishsurf (approved):
lgtm!

PR #537, opened 2021-09-24
chore: use bundle packaging format

flavianmissi:

PR #538, opened 2021-09-24
chore: use bundle packaging format

flavianmissi:
See https://github.com/quay/quay-operator/pull/537 for comments


bundle/downstream/manifests/quay-operator.clusterserviceversion.yaml
```diff
@@ -187,17 +187,17 @@ spec:
                     fieldRef:
                       fieldPath: metadata.annotations['quay-version']
                 - name: RELATED_IMAGE_COMPONENT_QUAY
-                  value: registry-proxy.engineering.redhat.com/rh-osbs/quay-quay-rhel8:v3.5.7
```
harishsurf, line 190: could we keep the redhat registry references as is since the CSV is for `downstream`. This makes it easier to fetch changes to dist-git during a downstream build. This is similar to what we have in `redhat-3.6` ([here](https://github.com/quay/quay-operator/blob/redhat-3.6/bundle/downstream/manifests/quay-operator.clusterserviceversion.yaml#L192))

harishsurf (approved):
left one comment - otherwise lgtm!

bundle/downstream/manifests/quay-operator.clusterserviceversion.yaml
ricardomaraschini: I had enough of `upstream` and `downstream` directories in this repo. We should have only one bundle and not even call it `upstream` at all. Any customization (`downstream`) should happen in the `downstream` repo (dist-git).

ricardomaraschini: And every body should be able to install from the `bundle` directory, so on this case we can't keep the `registry-proxy` thingy.

flavianmissi:
Merging this. @harishsurf we can talk more about the registry url if needed.

PR #539, opened 2021-09-27
migration: moving strategy to Recreate before upgrading (PROJQUAY-2586)

ricardomaraschini:
During upgrade from QuayEcosystem into QuayRegistry postgres deployment
is updated to include an extra env variable:

- POSTGRESQL_ADMIN_PASSWORD

Once the deployment is updated a new replica for the database is created
but it fails to roll out due to PVC not allowing multiple attachments.

This PR forces the deployment to adopt Recreate strategy as it is the
default in 3.6 postgres deployment.


ricardomaraschini:
@jonathankingfc this is an attempt to avoid the problem reported in the JIRA card. I have tested this manually by updating the deployment to use Recreate strategy before triggering the migration and it succeeded. Can you imagine in any downside on this approach?

jonathankingfc (approved):
LGTM

PR #541, opened 2021-09-28
upgrades: Fix CRD schema validation during upgrade (PROJQUAY-2587)

jonathankingfc:
- Remove unhealthyComponents as a required field to allow for pre-3.6 CRs to be valid during upgrade procedure
- Fix incorrect api group declarations


harishsurf (approved):
good catch! lgtm!

PR #542, opened 2021-09-29
database: avoid redeploy postgres during reconcile (PROJQUAY-2603)

ricardomaraschini:
Operator was setting an annotation in all created objects. This
annotation was not useful by anything except by the config tool.

This PR makes the annotation to be created only in the config tool
deployment.


flavianmissi (approved):
LGTM

jonathankingfc (approved):
LGTM

PR #544, opened 2021-09-29
status: omit conflict errors (PROJQUAY-2610)

ricardomaraschini:
These errors happen because we have two concurrent reconcilers operating
with the same resource. This is a design fault that need to be addressed
in the future. This PR omits conflict errors during status updates as
these are constantly retried.


jonathankingfc (approved):
LGTM

PR #546, opened 2021-09-29
tls: Remove tls certificates from reonciled secret (PROJQUAY-2606)

jonathankingfc:
- Prevent old tls certs from being included in new secret generated by controller

### DIFF
--- a/pkg/configure/configure.go
+++ b/pkg/configure/configure.go
@@ -73,22 +73,7 @@ func ReconfigureHandler(k8sClient client.Client) func(w http.ResponseWriter, r *
 			return
 		}
 
-		var oldSecret corev1.Secret
-		nsn.Name = quay.Spec.ConfigBundleSecret
-		if err := k8sClient.Get(context.Background(), nsn, &oldSecret); err != nil {
-			log.Error(
-				err,
-				"failed to fetch QuayRegistry config bundle",
-				"name",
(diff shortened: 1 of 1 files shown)


pkg/configure/configure.go
```diff
@@ -182,6 +182,9 @@ func createUpdatedSecret(reconfigureRequest request, oldSecret corev1.Secret) co
 	}
 
 	for prop, value := range oldSecret.Data {
+		if prop == "ssl.key" || prop == "ssl.cert" {
+			continue
+		}
```
ricardomaraschini: I don't think we need this `oldSecret` function argument anymore. I remember introducing this to solve https://issues.redhat.com/browse/PROJQUAY-2419 but it seems like after your fixes in the config editor it works without it. Can you remove it?

ricardomaraschini (changes requested):

pkg/configure/configure.go
ricardomaraschini: i.e. prior to your fix in the config editor it was not sending ssl.cert and ssl.key, hence this hack to inherit whatever was in the old secret.

jonathankingfc: Removed the logic for the old secret

ricardomaraschini (approved):

ricardomaraschini:
@jonathankingfc don't forget to cherry pick this into 3.6

PR #553, opened 2021-10-01
fix(workflow): use correct clair and builder-qemu tags

flavianmissi:
clair is in the process of removing their redhat-* branches, and asked
us to use their nightly instead.

I expected some builds to fail at first, but clair and builder-qemu are the exception - clair is removing their redhat-* branches, and the builder-qemu uses main instead of latest.
The rest of the expected tags will shortly be built manually so that we don't  depend on pushes to our various redhat-3.5 branches.

Hope this makes sense, hit me up on slack otherwise.


harishsurf (approved):
lgtm!

PR #554, opened 2021-10-01
migration: using edge route if tls type is none (PROJQUAY-2611)

ricardomaraschini:
If QuayEcosystem uses none as its TLS type we need to make sure TLS and
Route are managed in the created QuayRegistry.


jonathankingfc (approved):
lgtm

PR #556, opened 2021-10-01
refactor(build-workflow): infer image tag from branch name (PROJQUAY-2612)

flavianmissi:
This change ties the build workflow to the name of the branches it runs against.
In a future where downstream releases are fully automated, we should probably tag the redhat-* branches (for everything we want built, alpha, release candidates, etc). When we do this, this workflow should be updated to get triggered when new tags are created, so that it's more robust.

Successful workflow run: https://github.com/flavianmissi/quay-operator/actions/runs/1295760888


dmage (approved):
LGTM, just squash commits before merging them.

PR #564, opened 2021-10-05
componentstatus: using status.conditions for components (PROJQUAY-2636)

ricardomaraschini:
Reporting component status using .status.conditions instead of
unhealthyComponents property.


apis/quay/v1/quayregistry_types.go
```diff
@@ -92,19 +91,34 @@ type Component struct {
 
 type ConditionType string
 
+// Follow a list of known condition types. These are used when reporting a QuayRegistry status
+// through its .status.conditions slice.
 const (
 	ConditionTypeAvailable      ConditionType = "Available"
 	ConditionTypeRolloutBlocked ConditionType = "RolloutBlocked"
-	// TODO: Add more useful conditions.
```
flavianmissi, line 98: Beautiful.

controllers/quay/quayregistry_status_controller.go
```diff
@@ -123,212 +82,50 @@ func (q *QuayRegistryStatusReconciler) Reconcile(
...
-	}
-
-	conds := map[string][]qv1.Condition{}
-	for _, obc := range list.Items {
-		component, ok := obc.Annotations[quayComponentAnnotation]
-		if !qv1.Owns(reg, &obc) || !ok {
-			continue
+		// make sure we only update LastTransitionTime if the status has changed.
+		cond.LastTransitionTime = cond.LastUpdateTime
+		if curCond != nil && curCond.Status == cond.Status {
```
flavianmissi: AFAICS either the comment or the equality comparison here are wrong.
What I understand from the comment is that the `cond.LastTransitionTime` will only be updated if the status _changed_. 
The condition is however `curCond.Status == cond.Status`, which checks that the status has _not_ changed.
Did I misunderstanding something?

controllers/quay/quayregistry_status_controller.go
```diff
@@ -123,212 +82,50 @@ func (q *QuayRegistryStatusReconciler) Reconcile(
...
+			cond.LastTransitionTime = curCond.LastTransitionTime
 		}
 
-		phase := obc.Status.Phase
-		if phase == ocsv1a1.ObjectBucketClaimStatusPhaseBound {
-			continue
+		reg.Status.Conditions = qv1.SetCondition(reg.Status.Conditions, cond)
+		if cond.Status == metav1.ConditionFalse {
```
flavianmissi: We probably want to add a comment here expanding on the expectation that our conditions should be stated in a way that "True" means something works, and "False" means something does not work.

apis/quay/v1/quayregistry_types.go
```diff
@@ -92,19 +91,34 @@ type Component struct {
 
 type ConditionType string
 
+// Follow a list of known condition types. These are used when reporting a QuayRegistry status
+// through its .status.conditions slice.
```
flavianmissi, line 95: Same as below, I'll just copy/paste the comment.

We probably want to add a comment here expanding on the expectation that our conditions should be stated in a way that "True" means something works, and "False" means something does not work.

controllers/quay/quayregistry_status_controller.go
flavianmissi: I think I get it, the line just above this check defaults to updating the `LastUpdateTime`, but when the status hasn't changed we then set the time back to what it was.
Can we rephrase the comment maybe? I think that's what confused me.
A suggestion:
```
use original condition LastTransitionTime when the status hasn't changed
```

pkg/cmpstatus/base.go
```diff
@@ -0,0 +1,83 @@
...
+					Reason:         qv1.ConditionReasonComponentNotReady,
+					Message:        msg,
+					LastUpdateTime: metav1.NewTime(time.Now()),
+				}, nil
+			}
+			return zero, err
+		}
+
+		if !qv1.Owns(reg, &dep) {
+			msg := fmt.Sprintf("Deployment %s not found", depname)
```
flavianmissi: I wonder if it would be helpful to make this message a little more specific. It might be confusing for a user to see a deployment with the expected name while our condition is saying that it can't find it. Maybe we could say something in the lines of `"Deployment %s not owned by us"`.
wdyt?

pkg/cmpstatus/base.go
```diff
@@ -0,0 +1,83 @@
...
+
+		cond := b.deploy.check(dep)
+		if cond.Status != metav1.ConditionTrue {
+			// if the deployment is in a faulty state bails out immediately.
+			cond.Type = qv1.ComponentBaseReady
+			return cond, nil
+		}
+	}
+
+	return qv1.Condition{
```
flavianmissi, line 76: Why not use `cond` (returned by the call to `b.deploy.check`) here? We'd just need to set `cond.Type = qv1.ComponentBaseReady`, right?

flavianmissi (commented):
Good stuff! Left some minor comments.

controllers/quay/quayregistry_status_controller.go
ricardomaraschini: Yeah, this is indeed misleading. Good catch! I have refactored this.

controllers/quay/quayregistry_status_controller.go
ricardomaraschini: I think this would be much cleaner if we use
```go
reg.Status.Conditions = qv1.SetCondition(reg.Status.Conditions, cond)                                                                                                                                                     
if cond.Reason == qv1.ConditionReasonComponentNotReady {                                                                                                                                                                 
    faultySeen = true                                                                                                                                                                                                 
}
```
Good catch!

apis/quay/v1/quayregistry_types.go
ricardomaraschini: After the changes made in https://github.com/quay/quay-operator/pull/564#discussion_r732681091 I don't think this is necessary anymore. wdyt?

pkg/cmpstatus/base.go
ricardomaraschini: Makes total sense, will refactor this! Thanks for catching this, you may have saved us from some complains ;-)

pkg/cmpstatus/base.go
ricardomaraschini: Here we are checking a couple of deployments, if both of them are OK then we return this condition as a consolidated view of the other two. Not sure we need to change this, is that OK if we leave as is?

controllers/quay/quayregistry_status_controller.go
flavianmissi: huh I didn't think of using `cond.Reason` here, but yeah it's much more clear, good thinking!

apis/quay/v1/quayregistry_types.go
flavianmissi: Yup, agreed

pkg/cmpstatus/base.go
flavianmissi: Ah I didn't get that. It's totally ok to stay as is, thanks for expanding.

flavianmissi (approved):
LGTM. Great work! 🚀

PR #567, opened 2021-10-13
postgres: stop setting quay-registry-hostname annotation (PROJQUAY-2603)

ricardomaraschini:
This annotation is not useful for the postgres deployment in any
scenario.

### DIFF
--- a/pkg/middleware/middleware.go
+++ b/pkg/middleware/middleware.go
@@ -45,9 +45,20 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
 		return configBundleSecret, nil
 	}
 
-	// we have to set a special annotation in the config editor deployment. this annotation is
-	// then used as an environment variable and consumed by the app.
+	// we have to set a special annotation in the config editor deployment, this annotation is
+	// then used as an environment variable and consumed by the app. we also need to remove
+	// all unused annotations from postgres deployment to avoid its redeployment.
 	if dep, ok := obj.(*appsv1.Deployment); ok {
+		// TODO we must not set annotations in objects where they are not needed. we also
+		// must stop mangling objects in this "middleware" thingy, what is the point of
+		// using kustomize if we keep changing stuff on the fly ?
+		if strings.Contains(dep.GetName(), "quay-database") {
+			delete(dep.Spec.Template.Annotations, "quay-registry-hostname")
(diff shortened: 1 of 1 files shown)


pkg/middleware/middleware.go
```diff
@@ -45,9 +45,20 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
 
-	// we have to set a special annotation in the config editor deployment. this annotation is
-	// then used as an environment variable and consumed by the app.
+	// we have to set a special annotation in the config editor deployment, this annotation is
+	// then used as an environment variable and consumed by the app. we also need to remove
+	// all unused annotations from postgres deployment to avoid its redeployment.
 	if dep, ok := obj.(*appsv1.Deployment); ok {
+		// TODO we must not set annotations in objects where they are not needed. we also
+		// must stop mangling objects in this "middleware" thingy, what is the point of
+		// using kustomize if we keep changing stuff on the fly ?
```
flavianmissi, line 54: lol

pkg/middleware/middleware.go
```diff
@@ -45,9 +45,20 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+		if strings.Contains(dep.GetName(), "quay-database") {
+			delete(dep.Spec.Template.Annotations, "quay-registry-hostname")
+			delete(dep.Spec.Template.Annotations, "quay-buildmanager-hostname")
+			delete(dep.Spec.Template.Annotations, "quay-operator-service-endpoint")
+			return dep, nil
```
flavianmissi, line 59: I'm assuming you chose to delete these annotations here instead of not adding them to the postgres deployment in the first place because it is non-trivial to do so. Is that correct?

ricardomaraschini: Sort of, the reasoning is that the least we change at this stage the better.

flavianmissi: Got it. Do we track the need to stop adding these labels where they are not needed somewhere? Is there a JIRA?
I want us to get to this at some point, and from experience TODOs in the code tend to be forgotten/ignored (not that we can't do the same with JIRAs though).

flavianmissi (approved):

pkg/middleware/middleware.go
ricardomaraschini: You are right, I have created a card to keep track of it: https://issues.redhat.com/browse/PROJQUAY-2680

PR #571, opened 2021-10-19
upgradejob: decrease resource requests (PROJQUAY-2688)

ricardomaraschini:
Decreased resource.requests for the upgrade job. As it runs python
alembic the bulk of the load is in Postgres anyways. This pr does not
decrease the resource.limits for conservative purposes.

### DIFF
--- a/kustomize/base/upgrade.job.yaml
+++ b/kustomize/base/upgrade.job.yaml
@@ -51,8 +51,8 @@ spec:
               protocol: TCP
           resources:
             requests:
-              cpu: 2000m
-              memory: 8Gi
+              cpu: 1000m
+              memory: 3Gi
             limits:
               cpu: 2000m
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
After chatting with @kwestpharedhat  seems like 3Gi is a good amount of memory to request to run this migration. As we are not setting this as "limit" but rather "requests" this should be OK.

flavianmissi (approved):
LGTM 🎉 

Side question: are we backporting this to 3.6.1? Our e2e tests could use it.

ricardomaraschini:
@flavianmissi no, 3.6.1 is released, it can't be changed without generating 3.6.2.

ricardomaraschini:
@flavianmissi disregard my previous comment, we haven't released 3.6.1 yet. Will backport it.

PR #573, opened 2021-10-21
images: Enforce Postgres override image used in managed Clair (PROJQUAY-2122)

jonathankingfc:
- Use RELATED_IMAGE_COMPONENT_POSTGRES image override even when Postgres is unmanaged


ricardomaraschini (commented):
As we are here can you please remove the `panic` we have [here](https://github.com/quay/quay-operator/blob/master/pkg/kustomize/kustomize.go#L80) ? The way it is, it is very simple to render the operator out of work, one just needs to create an image overwrite with garbage.

ricardomaraschini (approved):
LGTM

PR #576, opened 2021-10-25
componentstatus: fix possible panic when setting LastTransitionTime (…

ricardomaraschini:
…PROJQUAY-2636)

We need to check if curCond is not nil before we move on and start using
its properties.


flavianmissi (approved):
Good catch!

ricardomaraschini:
> Good catch!

Kudos to @jonathankingfc

PR #582, opened 2021-11-02
chore: Increase the maximum number of PostgreSQL connections for Clair (PROJQUAY-2739)

ibazulic:
Previously, the number of PostgreSQL connections was set to 100 which is the default value. This value is too low and sometimes the pool gets exhausted pretty quickly. This PR will increase the number of max. connections to 1000 for the Clair PostgreSQL deployment which will allow new pods to connect.


jonathankingfc (approved):
LGTM

PR #586, opened 2021-11-04
volume: Add support for volume size and env variable overrides (PROJQUAY-1090)

jonathankingfc:
- Add overrides to component spec
- Add volumeSize to overrides spec field

### DIFF
--- a/apis/quay/v1/quayregistry_types.go
+++ b/apis/quay/v1/quayregistry_types.go
@@ -24,6 +24,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -38,15 +39,15 @@ type ComponentKind string
 
 const (
 	ComponentBase          ComponentKind = "base"
-	ComponentPostgres                    = "postgres"
-	ComponentClair                       = "clair"
-	ComponentRedis                       = "redis"
-	ComponentHPA                         = "horizontalpodautoscaler"
-	ComponentObjectStorage               = "objectstorage"
-	ComponentRoute                       = "route"
-	ComponentMirror                      = "mirror"
-	ComponentMonitoring                  = "monitoring"
-	ComponentTLS                         = "tls"
+	ComponentPostgres      ComponentKind = "postgres"
+	ComponentClair         ComponentKind = "clair"
+	ComponentRedis         ComponentKind = "redis"
+	ComponentHPA           ComponentKind = "horizontalpodautoscaler"
+	ComponentObjectStorage ComponentKind = "objectstorage"
+	ComponentRoute         ComponentKind = "route"
+	ComponentMirror        ComponentKind = "mirror"
+	ComponentMonitoring    ComponentKind = "monitoring"
+	ComponentTLS           ComponentKind = "tls"
 )
 
 var allComponents = []ComponentKind{
@@ -66,6 +67,11 @@ var requiredComponents = []ComponentKind{
 	ComponentObjectStorage,
 	ComponentRoute,
 	ComponentRedis,
+}
+
+var supportsVolumeOverride = []ComponentKind{
+	ComponentPostgres,
+	ComponentClair,
 }
 
 const (
@@ -88,6 +94,13 @@ type Component struct {
 	// Managed indicates whether or not the Operator is responsible for the lifecycle of this component.
 	// Default is true.
 	Managed bool `json:"managed"`
+	// Overrides holds information regarding component specific configurations.
+	Overrides *Override `json:"overrides,omitempty"`
+}
+
+// Override describes configuration overrides for the given managed component
+type Override struct {
+	VolumeSize *resource.Quantity `json:"volumeSize,omitempty"`
 }
 
 type ConditionType string
@@ -127,6 +140,7 @@ const (
 	ConditionReasonObjectStorageComponentDependencyError ConditionReason = "ObjectStorageComponentDependencyError"
 	ConditionReasonMonitoringComponentDependencyError    ConditionReason = "MonitoringComponentDependencyError"
 	ConditionReasonConfigInvalid                         ConditionReason = "ConfigInvalid"
+	ConditionReasonComponentOverrideInvalid              ConditionReason = "ComponentOverrideInvalid"
 )
 
 // Condition is a single condition of a QuayRegistry.
@@ -330,6 +344,33 @@ func EnsureDefaultComponents(ctx *quaycontext.QuayRegistryContext, quay *QuayReg
 	return updatedQuay, nil
 }
 
+// ValidateOverrides validates that the overrides set for each component are valid.
+func ValidateOverrides(quay *QuayRegistry) error {
+	for _, component := range quay.Spec.Components {
+
+		// No overrides provided
+		if component.Overrides == nil {
+			continue
+		}
+
+		// If the component is unmanaged, we cannot set overrides
+		if !ComponentIsManaged(quay.Spec.Components, component.Kind) {
+			if component.Overrides.VolumeSize != nil {
+				return errors.New("cannot set overrides on unmanaged component " + string(component.Kind))
+			}
+		}
+
+		// Check that component supports override
+		if component.Overrides.VolumeSize != nil && !ComponentSupportsOverride(component.Kind, "volumeSize") {
+			return fmt.Errorf("component %s does not support volumeSize overrides", component.Kind)
+		}
+
+	}
+
+	return nil
+
+}
+
 // EnsureRegistryEndpoint sets the `status.registryEndpoint` field and returns `ok` if it was unchanged.
 func EnsureRegistryEndpoint(ctx *quaycontext.QuayRegistryContext, quay *QuayRegistry, config map[string]interface{}) (*QuayRegistry, bool) {
 	updatedQuay := quay.DeepCopy()
@@ -491,6 +532,24 @@ func FieldGroupNamesForManagedComponents(quay *QuayRegistry) ([]string, error) {
 		fgns = append(fgns, fgn)
 	}
 	return fgns, nil
+}
+
+// ComponentSupportsOverride returns whether or not a given component supports the given override.
+func ComponentSupportsOverride(component ComponentKind, override string) bool {
+
+	// Using a switch statement for possible implementation of future overrides
+	switch override {
+	case "volumeSize":
+		for _, c := range supportsVolumeOverride {
+			if c == component {
+				return true
+			}
+		}
+	default:
+		return false
+	}
+
+	return false
 }
 
 func init() {
--- a/apis/quay/v1/quayregistry_types_test.go
+++ b/apis/quay/v1/quayregistry_types_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	quaycontext "github.com/quay/quay-operator/pkg/context"
@@ -349,6 +350,67 @@ var componentsMatchTests = []struct {
 		},
 		false,
 	},
+}
+
+var validateOverridesTests = []struct {
+	name        string
+	quay        QuayRegistry
+	expectedErr error
+}{
+	{
+		"NoOverridesProvided",
+		QuayRegistry{
+			Spec: QuayRegistrySpec{
+				Components: []Component{
+					{Kind: "postgres", Managed: true},
+					{Kind: "redis", Managed: true},
+					{Kind: "clair", Managed: true},
+					{Kind: "objectstorage", Managed: true},
+					{Kind: "route", Managed: true},
+					{Kind: "tls", Managed: true},
+					{Kind: "horizontalpodautoscaler", Managed: true},
+					{Kind: "mirror", Managed: true},
+					{Kind: "monitoring", Managed: true},
+				},
+			},
+		},
+		nil,
+	},
+	{
+		"InvalidVolumeSizeOverride",
+		QuayRegistry{
+			Spec: QuayRegistrySpec{
+				Components: []Component{
+					{Kind: "postgres", Managed: true},
+					{Kind: "redis", Managed: true},
+					{Kind: "clair", Managed: true},
+					{Kind: "objectstorage", Managed: true},
+					{Kind: "route", Managed: true},
+					{Kind: "tls", Managed: true, Overrides: &Override{VolumeSize: &resource.Quantity{}}},
+					{Kind: "horizontalpodautoscaler", Managed: true},
+					{Kind: "mirror", Managed: true},
+					{Kind: "monitoring", Managed: true},
+				},
+			},
+		},
+		errors.New("component tls does not support volumeSize overrides"),
+	},
+}
+
+func TestValidOverrides(t *testing.T) {
+	assert := assert.New(t)
+
+	for _, test := range validateOverridesTests {
+		t.Run(test.name, func(t *testing.T) {
+			err := ValidateOverrides(&test.quay)
+			if test.expectedErr != nil {
+				assert.NotNil(err, test.name)
+				assert.Equal(test.expectedErr, err)
+			} else {
+				assert.Equal(test.expectedErr, err)
+			}
+		})
+	}
 }
 
 func TestComponentsMatch(t *testing.T) {
--- a/pkg/middleware/middleware.go
+++ b/pkg/middleware/middleware.go
@@ -1,12 +1,14 @@
 package middleware
 
 import (
+	"fmt"
 	"strings"
 
 	route "github.com/openshift/api/route/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -60,7 +62,7 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
 		}
 
 		if !strings.Contains(dep.GetName(), "quay-config-editor") {
-			return obj, nil
+			return dep, nil
 		}
 
 		fgns, err := v1.FieldGroupNamesForManagedComponents(quay)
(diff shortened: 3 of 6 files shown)


dmesser:
I propose to add arbitrary labels as well. This will help with a lot of use cases where customers want to leverage labels to base automation and customization upon. One particular example is router sharding in OpenShift (https://docs.openshift.com/container-platform/4.9/networking/ingress-operator.html#nw-ingress-sharding-route-labels_configuring-ingress)

apis/quay/v1/quayregistry_types.go
```diff
@@ -88,6 +88,16 @@ type Component struct {
...
 	// Default is true.
 	Managed bool `json:"managed"`
+	// Overrides holds information regarding component specific configurations.
+	Overrides Override `json:"overrides,omitempty"`
+}
+
+// Override describes configuration overrides for the given managed component
+type Override struct {
+	// Env holds environment variable overrides that will be added to the given managed components deployment (if available)
+	Env []string `json:"env,omitempty"`
```
flavianmissi: We could make this a struct so we limit what env vars can be exported to services.
Wdyt?

flavianmissi: Talked to @dmesser about this and for now it's better for us to stay flexible so we cover the "unknown unknowns".
With that said, shouldn't `Env` be a `map[string]string` ?

apis/quay/v1/quayregistry_types.go
```diff
@@ -37,16 +37,16 @@ var QuayVersionCurrent QuayVersion = QuayVersion(os.Getenv("QUAY_VERSION"))
...
-	ComponentPostgres                    = "postgres"
-	ComponentClair                       = "clair"
-	ComponentRedis                       = "redis"
-	ComponentHPA                         = "horizontalpodautoscaler"
-	ComponentObjectStorage               = "objectstorage"
-	ComponentRoute                       = "route"
-	ComponentMirror                      = "mirror"
-	ComponentMonitoring                  = "monitoring"
-	ComponentTLS                         = "tls"
+	ComponentBase          = "base"
```
dmage: I would keep these constants typed.

ricardomaraschini: The way it was only the ComponentBase had a different type, all the others were strings: https://goplay.space/#C-DHpg_1hcO

apis/quay/v1/quayregistry_types.go
```diff
@@ -68,6 +68,19 @@ var requiredComponents = []ComponentKind{
 	ComponentRedis,
 }
 
+// fixme (jonathankingfc) - These should not be strings, use a correct type
```
ricardomaraschini: We must not have "fixme" things if we gonna make this part of a release.

apis/quay/v1/quayregistry_types.go
```diff
@@ -68,6 +68,19 @@ var requiredComponents = []ComponentKind{
...
+var overrideSupport = map[ComponentKind][]string{
```
ricardomaraschini: For sake of sanity can you make all constants of type "ComponentKind'' and not string.

apis/quay/v1/quayregistry_types.go
```diff
@@ -127,6 +152,7 @@ const (
 	ConditionReasonObjectStorageComponentDependencyError ConditionReason = "ObjectStorageComponentDependencyError"
 	ConditionReasonMonitoringComponentDependencyError    ConditionReason = "MonitoringComponentDependencyError"
 	ConditionReasonConfigInvalid                         ConditionReason = "ConfigInvalid"
+	ConditionReasonInvalidComponentOverride              ConditionReason = "ComponentOverrideInvalid"
```
ricardomaraschini: InvalidComponentOverride or ComponnetOverrideInvalid, can we go with only one of them?

apis/quay/v1/quayregistry_types.go
```diff
@@ -330,6 +356,34 @@ func EnsureDefaultComponents(ctx *quaycontext.QuayRegistryContext, quay *QuayReg
 	return updatedQuay, nil
 }
 
+// ValidateOverrides validates that the overrides set for each component are valid.
+func ValidateOverrides(quay *QuayRegistry) error {
+	for _, component := range quay.Spec.Components {
+
+		// If the component is unmanaged, we cannot set overrides
+		if !ComponentIsManaged(quay.Spec.Components, component.Kind) {
+			if component.Overrides.Env != nil || component.Overrides.Labels != nil || component.Overrides.VolumeSize != "" {
```
ricardomaraschini: What about making this a method in "Override" ? We could use here something like "override.IsEmpty()"?

apis/quay/v1/quayregistry_types.go
```diff
@@ -330,6 +356,34 @@ func EnsureDefaultComponents(ctx *quaycontext.QuayRegistryContext, quay *QuayReg
...
+				return errors.New("cannot set overrides on unmanaged component " + string(component.Kind))
+			}
```
ricardomaraschini, line 360: Here we could also "continue" as that is all we need to verify for an "unmanaged component"

apis/quay/v1/quayregistry_types.go
```diff
@@ -330,6 +356,34 @@ func EnsureDefaultComponents(ctx *quaycontext.QuayRegistryContext, quay *QuayReg
...
+		}
+
+		// Validate fields
+		if component.Overrides.Env != nil && !ComponentSupportsOverride(component.Kind, "env") {
```
ricardomaraschini: `nil` and an empty slice are different things, you should use `len(slice) != 0` here.

apis/quay/v1/quayregistry_types.go
```diff
@@ -493,6 +547,15 @@ func FieldGroupNamesForManagedComponents(quay *QuayRegistry) ([]string, error) {
 	return fgns, nil
 }
 
+func ComponentSupportsOverride(component ComponentKind, override string) bool {
```
ricardomaraschini, line 538: Can you add a comment here? Linter is complaining. Also if one passes "ComponentBase" here this will "panic".

apis/quay/v1/quayregistry_types_test.go
```diff
@@ -351,6 +351,115 @@ var componentsMatchTests = []struct {
...
+}
+
+func TestValidOverrides(t *testing.T) {
+	assert := assert.New(t)
+
+	for _, test := range validateOverridesTests {
+		err := ValidateOverrides(&test.quay)
+
+		if test.expectedErr != nil {
+			assert.NotNil(err, test.name)
```
ricardomaraschini: test.name will never be nil.

apis/quay/v1/quayregistry_types_test.go
```diff
@@ -351,6 +351,115 @@ var componentsMatchTests = []struct {
...
+			},
+		},
+		errors.New("cannot set overrides on unmanaged component postgres"),
+	},
+}
+
+func TestValidOverrides(t *testing.T) {
+	assert := assert.New(t)
+
+	for _, test := range validateOverridesTests {
```
ricardomaraschini, line 403: By using t.Run() here and passing in the test name as argument makes the test output easier to read.

apis/quay/v1/quayregistry_types.go
```diff
@@ -68,6 +68,19 @@ var requiredComponents = []ComponentKind{
...
+var overrideSupport = map[ComponentKind][]string{
+	ComponentPostgres:      {"volumeSize", "labels", "env"},
+	ComponentClair:         {"volumeSize", "labels", "env"},
+	ComponentRedis:         {"labels", "env"},
+	ComponentHPA:           {"labels"},
+	ComponentObjectStorage: {"volumeSize", "labels", "env"},
+	ComponentRoute:         {"labels"},
+	ComponentMirror:        {"labels", "env"},
+	ComponentMonitoring:    {"labels"},
+	ComponentTLS:           {"labels"},
```
ricardomaraschini: Why TLS supports labels? Where are these labels set?

pkg/middleware/middleware.go
```diff
@@ -59,8 +60,22 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
 			return dep, nil
 		}
 
+		// Set env variable overrides
```
ricardomaraschini: You are not setting env variables for the "quay-database" deployment (there is a return on line 60 above). Is that on purpose?

pkg/middleware/middleware.go
```diff
@@ -59,8 +60,22 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+		var envVars map[string]string
+		switch quayComponentLabel {
+		case "postgres":
+			envVars = getEnvOverrideForComponent(quay, v1.ComponentPostgres)
+		case "clair-postgres":
+			envVars = getEnvOverrideForComponent(quay, v1.ComponentClair)
+		case "objectstorage":
```
ricardomaraschini: What deployment contains as its component label "objectstorage" ? This does not seem right. where are we setting the override for the object storage?

pkg/middleware/middleware.go
```diff
@@ -72,6 +87,21 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
 
+	// If the current object is a PVC, check for volume override
+	if pvc, ok := obj.(*corev1.PersistentVolumeClaim); ok {
+		var volumeSize resource.Quantity
+		switch quayComponentLabel {
+		case "postgres":
+			volumeSize = getVolumeSizeOverrideForComponent(quay, v1.ComponentPostgres)
+		case "clair-postgres":
+			volumeSize = getVolumeSizeOverrideForComponent(quay, v1.ComponentClair)
+		case "objectstorage":
```
ricardomaraschini: I can't see this working. Do we create a persistent volume claim for the object storage? I thought we were only creating an ObjectBucketClaim. Can you clear this for me?

pkg/middleware/middleware.go
```diff
@@ -72,6 +87,21 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+			volumeSize = getVolumeSizeOverrideForComponent(quay, v1.ComponentObjectStorage)
+		}
+		pvc.Spec.Resources.Requests = corev1.ResourceList{corev1.ResourceStorage: volumeSize}
```
ricardomaraschini: What happens when users shrink the storage? Can this operation generate some failure? If yes, does the user see the failure in the quay registry object status?

ricardomaraschini (changes requested):
Please take a look at my comments.

apis/quay/v1/quayregistry_types.go
```diff
@@ -88,6 +101,18 @@ type Component struct {
...
 	// Default is true.
 	Managed bool `json:"managed"`
+	// Overrides holds information regarding component specific configurations.
+	Overrides Override `json:"overrides,omitempty"`
+}
+
+// Override describes configuration overrides for the given managed component
+type Override struct {
+	// Env holds environment variable overrides that will be added to the given managed components deployment (if available)
+	Env map[string]string `json:"env,omitempty"`
```
ricardomaraschini: I think we might consider using [EnvVar](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L1981-L2001) type here. WDYT?

pkg/middleware/middleware.go
```diff
@@ -59,8 +60,22 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+		switch quayComponentLabel {
+		case "postgres":
+			envVars = getEnvOverrideForComponent(quay, v1.ComponentPostgres)
+		case "clair-postgres":
+			envVars = getEnvOverrideForComponent(quay, v1.ComponentClair)
+		case "objectstorage":
+			envVars = getEnvOverrideForComponent(quay, v1.ComponentObjectStorage)
+		}
+		for key, val := range envVars {
+			dep.Spec.Template.Spec.Containers[0].Env = append(dep.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: key, Value: val})
```
ricardomaraschini: This will set only in the first container. Is this expected? Why not set in all of them?

flavianmissi (commented):
Hey @jonathankingfc,
I thought we had agreed that we would focus on overrides, but this PR includes both labels and env vars.
What are you plans here?

apis/quay/v1/quayregistry_types.go
jonathankingfc: Changed back to Componentkind type

apis/quay/v1/quayregistry_types.go
jonathankingfc: Removed this implementation. Using an array of ComponentKind instead

apis/quay/v1/quayregistry_types.go
jonathankingfc: Removed label implementation for this PR

apis/quay/v1/quayregistry_types.go
jonathankingfc: Removed env implementation for this PR

apis/quay/v1/quayregistry_types.go
jonathankingfc: Resolved

jonathankingfc:
> Hey @jonathankingfc, I thought we had agreed that we would focus on overrides, but this PR includes both labels and env vars. What are you plans here?

@flavianmissi Updated the PR with fixes proposed by @ricardomaraschini and removed any implementation of env and label overrides (this should be supported in a later PR).

apis/quay/v1/quayregistry_types.go
ricardomaraschini: I think you gonna have to add `ComponentKind` to each line, otherwise only the first one is of type `ComponentKind`. Can you do that?

pkg/middleware/middleware.go
ricardomaraschini: @jonathankingfc I still don't understand this. What is this `case` supposed to do?

pkg/middleware/middleware.go
ricardomaraschini: @jonathankingfc ping

apis/quay/v1/quayregistry_types.go
```diff
@@ -493,6 +528,24 @@ func FieldGroupNamesForManagedComponents(quay *QuayRegistry) ([]string, error) {
 	return fgns, nil
 }
 
+// ComponentSupportsOverride returns whether or not a given component supports the given override.
+func ComponentSupportsOverride(component ComponentKind, override string) bool {
+
+	// Using a swtich statement for possible implementation of future overrides
```
ricardomaraschini: small typo here in "switch"

pkg/middleware/middleware.go
```diff
@@ -137,3 +157,15 @@ func FlattenSecret(configBundle *corev1.Secret) (*corev1.Secret, error) {
 
 	return flattenedSecret, nil
 }
+
+func getVolumeSizeOverrideForComponent(quay *v1.QuayRegistry, componentKind v1.ComponentKind) resource.Quantity {
+	for _, component := range quay.Spec.Components {
+		if component.Kind == componentKind {
+			volumeSizeOverride := component.Overrides.VolumeSize
+			if volumeSizeOverride != "" {
+				return resource.MustParse(volumeSizeOverride)
```
ricardomaraschini: This is wrong. We should not panic if user enters with wrong values.

pkg/middleware/middleware.go
jonathankingfc: This is removed from the PR.

pkg/middleware/middleware.go
jonathankingfc: This switches on the label taken from the current resource in the middleware pipeline. We could move this logic into the getVolumeSizeOverride() function, though

ricardomaraschini: My question is because I don't think we have a PVC with such label ("objectstorage"), hence my confusion I was wishing to get straight. Who creates this PVC ? As far as I remember we only have two PVCs (quay and clair databases).

flavianmissi: I've seen mentions to this component in the logs. From what I gather the operator doesn't currently have the option of managing it, but it's still treated as a component.
@jonathankingfc please confirm.

jonathankingfc: @flavianmissi since the PVC for objectstorage is created manually by the user during the pre-install process (we want to keep this behavior), we will not be touching the object storage PVC. I have removed this line from the PR

apis/quay/v1/quayregistry_types.go
```diff
@@ -68,6 +68,12 @@ var requiredComponents = []ComponentKind{
 	ComponentRedis,
 }
 
+var supportsVolumeOverride = []ComponentKind{
+	ComponentPostgres,
+	ComponentClair,
+	ComponentObjectStorage,
```
ricardomaraschini: Does ObjectStorage support Override ? The Jira card has, as an acceptance criteria:

"volume overrides available for every managed component that uses a PersistentVolume (**Noobaa Backend**, Postgres for Quay, Postgres for Clair)" 

I  know that we do not set the size of the Nooba Backend but before merging this we better clarify the behavior with the Jira card reporter.

pkg/middleware/middleware.go
```diff
@@ -72,6 +74,28 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+			if err != nil {
+				return nil, fmt.Errorf("could not override volume size: " + err.Error())
+			}
+		case "clair-postgres":
+			volumeSizeOverride, err = getVolumeSizeOverrideForComponent(quay, v1.ComponentClair)
+			if err != nil {
+				return nil, fmt.Errorf("could not override volume size: " + err.Error())
+			}
+		}
+		if volumeSizeOverride.Cmp(*pvc.Spec.Resources.Requests.Storage()) == -1 {
```
ricardomaraschini: We shouldn't blindly trust that we can dereference the return of `Storage()`. Can you check if the returned value is not nil before calling `Cmp()` ? In this scenario I guess it is OK but it does not hurt to make this more "defensive".

apis/quay/v1/quayregistry_types.go
```diff
@@ -88,6 +94,13 @@ type Component struct {
 	// Managed indicates whether or not the Operator is responsible for the lifecycle of this component.
 	// Default is true.
 	Managed bool `json:"managed"`
+	// Overrides holds information regarding component specific configurations.
+	Overrides Override `json:"overrides,omitempty"`
```
ricardomaraschini: Have you seen how this show up in the yaml when empty? Something like `oc edit quayregistry/name` should suffice to make sure this is not showing up empty when `VolumeSize` is not set. I am asking because omitempty, in most of the cases, still shows empty structs (except when it is a pointer to a struct and it is nil).

apis/quay/v1/quayregistry_types.go
```diff
@@ -88,6 +94,13 @@ type Component struct {
...
+}
+
+// Override describes configuration overrides for the given managed component
+type Override struct {
+	VolumeSize string `json:"volumeSize,omitempty"`
```
ricardomaraschini: This could be of type `*resource.Quantity`, wdyt?

pkg/middleware/middleware.go
```diff
@@ -72,6 +74,28 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+			return nil, fmt.Errorf("cannot shrink volume override size from %s to %s", volumeSizeOverride.String(), pvc.Spec.Resources.Requests.Storage().String())
```
ricardomaraschini: We shouldn't blindly trust that we can dereference the return of `Storage()` here. Another thing is that this makes us not to support volume shrink, I guess this is OK but is the Jira card reporter aware of this? Can you discuss this and keep the card updated ? (makes QE and our life easier down the line).

I have noticed that storage resizes for mounted volumes only occur when the pod is restarted (at least for gp2 storage classes), remaining with a `FileSystemResizePending` condition where users can read: 

```Waiting for user to (re-)start a pod to finish file system resize of volume on node.```

It remains in this state until the pod where it is mounted is restarted. How are we restarting postgres pods ? Does k8s restarts them automatically after a while? Shouldn't we take this condition into account when reporting statuses for `postgres` and `clair` components?

**Review by ricardomaraschini — CHANGES_REQUESTED** [review]
Almost there, just clarifying some things below.

File: pkg/middleware/middleware.go (hunk 25 above)
Comment (jonathankingfc) [reply pkg/middleware/middleware.go]: I've added a nil check for the Storage() value just in case, however in this scenario it would never be nil anyway. Also, we cannot support volume shrinking as this is not even supported by most, if not all, storage classes. Also, the PVCs are created during the re-reconcile (which would be triggered by setting an override). The pods are also redeployed in that reconcile loop and will mount the new volume.

File: pkg/middleware/middleware.go [hunk 26]
```diff
@@ -72,6 +74,28 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
...
+			volumeSizeOverride = getVolumeSizeOverrideForComponent(quay, v1.ComponentPostgres)
+		case "clair-postgres":
+			volumeSizeOverride = getVolumeSizeOverrideForComponent(quay, v1.ComponentClair)
+		}
+
+		// If override was provided
+		if volumeSizeOverride != nil {
+			// Ensure that volume size is not being reduced
+			if pvc.Spec.Resources.Requests.Storage() != nil && volumeSizeOverride.Cmp(*pvc.Spec.Resources.Requests.Storage()) == -1 {
+				return nil, fmt.Errorf("cannot shrink volume override size from %s to %s", volumeSizeOverride.String(), pvc.Spec.Resources.Requests.Storage().String())
```
ricardomaraschini: Aren't `from` and `to` inverted here?

ricardomaraschini (commented):
All good! Let's just await for @dmesser in the Jira card then we can get this in.

pkg/middleware/middleware.go
ricardomaraschini: @jonathankingfc this is the last one, I promise :-)

jonathankingfc: Fixed!

ricardomaraschini (approved):

PR #590, opened 2021-11-25
hack: setup and teardown scripts for e2e envs

flavianmissi:
includes 3 shell scripts:
- `hack/deploy.sh`: deploys the operator and a quayregistry instance, and ensures everything is up and running. will exit 1 otherwise
- `hack/teardown.sh`: removes everything created by `deploy.sh`
- `hack/storage.sh`: installs the openshift container storage from the red hat catalog and creates a noobaa instance

note that some functions, like `info`, are duplicated in different files. since they are trivial, I didn't see much of a problem with it, at least for now.

Fixes PROJQUAY-2823

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -1,0 +1,223 @@
+#!/bin/bash
+# deploys the operator to a cluster from a published version of its catalog
+# index.
+#
+# by default, the deployed operator image is
+# `quay.io/projectquay/quay-operator-index:3.6-unstable`. you can override the
+# tag alone by exporting TAG before exeucting this script, or override the
+# whole thing by setting CATALOG_IMAGE.
+#
+# REQUIREMENTS:
+#  * a valid login session to an OCP cluster, with cluster admin privileges
+#  * the namespace $NAMESPACE exists in the cluster
+#  * noobaa backing storage (TODO: elaborate)
+#  * `curl`
+#  * `docker`
+#  * `yq`
+#  * `jq`
+#  * `oc`
+#
+# NOTE: this script will modify the following files:
+#  - bundle/quay-operator.catalogsource.yaml
+#  - bundle/quay-operator.operator-group.yaml
+#  - bundle/quay-operator.subscription.yaml
+# if `git` is available it will be used to checkout changes to the above files.
+# this means that if you made any changes to them and want them to be persisted,
+# make sure to commit them before running this script.
+
+# prints pre-formatted info output.
+function info {
+	echo "INFO $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
+
+# prints pre-formatted error output.
+function error {
+	>&2 echo "ERROR $(date '+%Y-%m-%dT%H:%M:%S') $*"
+}
+
+export TAG=${TAG:-'3.6-unstable'}
+export CATALOG_PATH=${CATALOG_PATH:-'./bundle/quay-operator.catalogsource.yaml'}
+export CATALOG_IMAGE=${CATALOG_IMAGE:-"quay.io/projectquay/quay-operator-index:${TAG}"}
+export OG_PATH=${OG_PATH:-'./bundle/quay-operator.operatorgroup.yaml'}
+export SUBSCRIPTION_PATH=${SUBSCRIPTION_PATH:-'./bundle/quay-operator.subscription.yaml'}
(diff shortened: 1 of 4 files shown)


dmage (dismissed):

hack/deploy.sh
```diff
@@ -0,0 +1,205 @@
...
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-config-editor --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-mirror --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-app --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+
+
+# manually check quay's health to ensure it can successfuly connect to its components.
+endpoint=`oc get quayregistry skynet -o jsonpath='{.status.registryEndpoint}'`
+result=`curl -s -k "${endpoint}"/health/instance`
+
+if [ ! `echo "${result}"|jq '.status_code'` = "200" ]; then
```
dmage:
```suggestion
if [ "$(echo "$result" | jq '.status_code')" != "200" ]; then
```

flavianmissi: Good catch!
I've applied the change to this and all following conditions, so I won't commit the suggestion via UI.

flavianmissi:
e2e tests have started failing after some point in this PR (https://github.com/flavianmissi/quay-operator/runs/4332403123?check_suite_focus=true).
Please hold approaval until I have figured out what's up.

flavianmissi:
Problem was that the line below, in `hack/deploy.sh` was missing the namespace, so it worked locally but not in the CI.
```bash
endpoint="$(oc get quayregistry skynet -o jsonpath='{.status.registryEndpoint}')"
```
After adding the namespace the new run is successful: https://github.com/flavianmissi/quay-operator/runs/4332862425?check_suite_focus=true
@dmage this is one is ready to go now, please have another look at your earliest convenience 🚀

hack/deploy.sh
```diff
@@ -0,0 +1,221 @@
...
+
+# manually check quay's health to ensure it can successfuly connect to its components.
+endpoint="$(oc -n "${NAMESPACE}" get quayregistry skynet -o jsonpath='{.status.registryEndpoint}')"
+result="$(curl -s -k "${endpoint}"/health/instance)"
+
+if [ "$(echo "${result}" | jq '.status_code')" != "200" ]; then
+	error 'quay health check did not return 200'
+	info "${result}"
+	info 'quay-app pod yaml:'
+	oc -n "${NAMESPACE}" logs -l=quay-component=quay-app
```
ricardomaraschini, line 169: can we dump here the quayregistry object as well?

hack/deploy.sh
```diff
@@ -0,0 +1,221 @@
...
+oc -n "${NAMESPACE}" wait pods -l=quay-component=redis --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+
+info 'waiting for mirror pod...'
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-mirror --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+
+info 'waiting for config-editor pod...'
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-config-editor --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
+
+info 'waiting for quay pod...'
+oc -n "${NAMESPACE}" wait pods -l=quay-component=quay-app --for=condition=Ready --timeout="${WAIT_TIMEOUT}"
```
ricardomaraschini, line 156: what do you think about checking quayregistry conditions instead of each component individually?

flavianmissi: Yeah that's a good idea. Can we still use `oc wait` to wait for them? Additionally, what condition would we wait for?

hack/deploy.sh
flavianmissi: Good call, will do!

flavianmissi: done - see two lines above

ricardomaraschini (approved):
/lgtm

hack/deploy.sh
flavianmissi: Talked about it briefly on slack - let's address this on a follow up PR

PR #591, opened 2021-11-26
refactor: nuke goroutine polling migrations

flavianmissi:
also introduces a new condition reason for migration failure.

fixes [PROJQUAY-1812](https://issues.redhat.com/browse/PROJQUAY-1812)

---

Test cases (please help me think of more):

* first installation successfully runs the upgrade job and install quay (I have tested this and it works)

  * from inside the repository root (with this branch cloned):
  * start the operator with `QUAY_VERSION='dev' go run main.go`
  * create the RD with `oc create -f config/samples/managed.quayregistry.yaml`
  * `watch oc get pods`
  * operator will successfully install and start Quay. there should also be one upgrade job completed.

* upgrade existing installation (I have not tried this)
  - Would it require building a catalog with say version 3.6.2-alpha, installing 3.6.1 via startingCSV, then upgrade to 3.6.2-alpha (which will have the code from this PR)?
  - Or is there another easier way?

* Update database config setting wrong password, and manually apply config changes
  * I did this by manually POSTing a request to the config editor
  * You can use this script to do it like I did: https://gist.github.com/flavianmissi/89d2891612fbb009ad677fa31e945d2e
    * Note that there is a manual step required - editing the DB_URI to set the wrong password, as specified on line 26 of the gist
  * Once the operator reconciles the changes, the upgrade job will error
  * Update the config again, setting the correct password back this time
  * The operator will recreate the upgrade job, and this time it will succeed

Open questions:
* how is QUAY_VERSION populated in non-dev deployments of the operator?

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -120,10 +120,103 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
-	created := v1.GetCondition(quay.Status.Conditions, v1.ConditionComponentsCreated)
-	if created != nil && created.Reason == v1.ConditionReasonMigrationsInProgress {
-		log.Info("migrations in progress, skipping reconcile")
-		return ctrl.Result{}, nil
+	created := v1.GetCondition(updatedQuay.Status.Conditions, v1.ConditionComponentsCreated)
+	migrationInProgress := created != nil && created.Reason == v1.ConditionReasonMigrationsInProgress
+	if migrationInProgress {
+		log.Info("checking Quay upgrade `Job` completion")
+
+		var upgradeJob batchv1.Job
+		err := r.Client.Get(
+			ctx,
+			types.NamespacedName{
+				Name:      updatedQuay.GetName() + "-" + v1.QuayUpgradeJobName,
+				Namespace: updatedQuay.GetNamespace(),
+			},
+			&upgradeJob,
+		)
+
+		// similarly to when a v1.ConditionReasonMigrationsFailed occurs,
+		// when the upgrade job is expected to exist but doesn't
+		// (i.e. someone manually removed it) we want the reconcile loop
+		// to run in its entirety, so we change the condition reason to
+		// something other than migrations in progress.
+		if err != nil && errors.IsNotFound(err) {
+			_, err = r.updateWithCondition(
+				updatedQuay,
+				v1.ConditionComponentsCreated,
+				metav1.ConditionFalse,
+				v1.ConditionReasonMigrationsJobMissing,
+				"upgrade job not found")
+			if err != nil {
+				log.Error(err, "failed to update `conditions` of `QuayRegistry`")
+			}
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if err != nil {
+			log.Error(err, "could not retrieve Quay upgrade `Job`")
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if upgradeJob.Status.Active == 1 {
+			log.Info("Upgrade job running, requeueing reconcile...")
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if upgradeJob.Status.Succeeded == 1 {
+			log.Info("Quay upgrade complete, updating `status.currentVersion`")
+
+			msg := "All registry components created"
+			condition := v1.Condition{
+				Type:               v1.ConditionComponentsCreated,
+				Status:             metav1.ConditionTrue,
+				Reason:             v1.ConditionReasonComponentsCreationSuccess,
+				Message:            msg,
+				LastUpdateTime:     metav1.Now(),
+				LastTransitionTime: metav1.Now(),
+			}
+			updatedQuay.Status.Conditions = v1.SetCondition(updatedQuay.Status.Conditions, condition)
+			updatedQuay.Status.CurrentVersion = v1.QuayVersionCurrent
+
+			if err = r.Client.Status().Update(ctx, updatedQuay); err != nil {
+				log.Error(err, "could not update QuayRegistry status with current version")
+				return ctrl.Result{RequeueAfter: time.Second}, nil
+			}
+
+			log.Info("successfully updated `status` after Quay upgrade")
+
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		// when the job fails, the next reconciliation to run will think
+		// that migrations are currently not running, since we change
+		// the condition reason to v1.ConditionReasonMigrationsFailed.
+		// this doesn't necessarily describe reality, because kube will
+		// retry failed jobs for us, but it's desired behaviour because
+		// when the migration job fails due to misconfiguration, then the
+		// reconcile function should be allowed to proceed.
+		if upgradeJob.Status.Failed == 1 {
+			msg := "failed to run migrations"
+			for _, cond := range upgradeJob.Status.Conditions {
+				if cond.Type == batchv1.JobFailed && cond.Status == corev1.ConditionTrue {
+					msg = cond.Message
+					break
+				}
+			}
+			updatedQuay, err = r.updateWithCondition(
+				updatedQuay,
+				v1.ConditionComponentsCreated,
+				metav1.ConditionFalse,
+				v1.ConditionReasonMigrationsFailed,
+				msg)
+			if err != nil {
+				log.Error(err, "failed to update `conditions` of `QuayRegistry`")
+			}
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		log.Info("Unexpected job status.", fmt.Sprintf("%+v", upgradeJob.Status))
+		return ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
 	if quay.Spec.ConfigBundleSecret == "" {
@@ -352,68 +445,8 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		updatedQuay, err = r.updateWithCondition(updatedQuay, v1.ConditionComponentsCreated, metav1.ConditionFalse, v1.ConditionReasonMigrationsInProgress, "running database migrations")
 		if err != nil {
 			log.Error(err, "failed to update `conditions` of `QuayRegistry`")
-
-			return ctrl.Result{}, nil
 		}
-
-		go func(quayRegistry *v1.QuayRegistry) {
-			err = wait.Poll(upgradePollInterval, upgradePollTimeout, func() (bool, error) {
-				log.Info("checking Quay upgrade `Job` completion")
-
-				var upgradeJob batchv1.Job
-				err = r.Client.Get(ctx, types.NamespacedName{Name: quayRegistry.GetName() + "-quay-app-upgrade", Namespace: quayRegistry.GetNamespace()}, &upgradeJob)
-				if err != nil {
-					log.Error(err, "could not retrieve Quay upgrade `Job`")
-
-					return false, err
-				}
-
-				if upgradeJob.Status.Succeeded > 0 {
-					log.Info("Quay upgrade complete, updating `status.currentVersion`")
-
-					var freshQuay v1.QuayRegistry
-					if err := r.Client.Get(ctx, req.NamespacedName, &freshQuay); err != nil {
-						log.Error(err, "could not retrieve QuayRegistry")
-						return false, err
-					}
-					qcopy := freshQuay.DeepCopy()
-
-					msg := "All registry components created"
-					condition := v1.Condition{
-						Type:               v1.ConditionComponentsCreated,
-						Status:             metav1.ConditionTrue,
-						Reason:             v1.ConditionReasonComponentsCreationSuccess,
-						Message:            msg,
-						LastUpdateTime:     metav1.Now(),
-						LastTransitionTime: metav1.Now(),
-					}
-					qcopy.Status.Conditions = v1.SetCondition(qcopy.Status.Conditions, condition)
-					qcopy.Status.CurrentVersion = v1.QuayVersionCurrent
-					r.EventRecorder.Event(qcopy, corev1.EventTypeNormal, string(v1.ConditionReasonHealthChecksPassing), msg)
-
-					if err = r.Client.Status().Update(ctx, qcopy); err != nil {
-						log.Error(err, "could not update QuayRegistry status with current version")
-						return false, nil
-					}
-
-					qcopy.Spec.Components = v1.EnsureComponents(qcopy.Spec.Components)
-					if err = r.Client.Update(ctx, qcopy); err != nil {
-						log.Error(err, "could not update QuayRegistry spec to complete upgrade")
-						return false, nil
-					}
-
-					log.Info("successfully updated `status` after Quay upgrade")
-
-					return true, nil
-				}
-
-				return false, nil
-			})
-
-			if err != nil {
-				log.Error(err, "Quay upgrade `Job` never completed")
-			}
-		}(updatedQuay.DeepCopy())
+		return ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
 	if !controllerutil.ContainsFinalizer(updatedQuay, QuayOperatorFinalizer) {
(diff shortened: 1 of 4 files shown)


pkg/kustomize/kustomize.go
```diff
@@ -492,16 +492,24 @@ func Inflate(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, baseCo
...
+	migrationInProgress := created != nil && created.Reason == v1.ConditionReasonMigrationsInProgress
+	log.Info("----------------------------------------------------------------------------------")
+	log.Info(fmt.Sprintf("migrationInProgress is: '%t'", migrationInProgress))
+	log.Info(fmt.Sprintf("currentVersion is: '%s'", quay.Status.CurrentVersion))
+
 	var overlay string
 	if rolloutBlocked(quay) {
 		overlay = configEditorOnlyOverlay()
-	} else if quay.Status.CurrentVersion != v1.QuayVersionCurrent {
+	} else if quay.Status.CurrentVersion != v1.QuayVersionCurrent && !migrationInProgress {
```
ricardomaraschini: I guess it is safe to assume that this part of the code won't be executed if the migrations are running (ConditionReasonMigrationsInProgress). No need to check for `!migrationInProgress`.

controllers/quay/quayregistry_controller.go
```diff
@@ -345,68 +419,8 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-				}
-
-				return false, nil
-			})
-
-			if err != nil {
-				log.Error(err, "Quay upgrade `Job` never completed")
-			}
-		}(updatedQuay.DeepCopy())
+		return ctrl.Result{RequeueAfter: time.Millisecond * 1000}, nil
```
ricardomaraschini: Isn't `time.Millisecond * 1000` equal to `time.Second` ? I have seen this in different places across this codebase, I prefer the latter.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,84 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		log.Info("migrations in progress, skipping reconcile")
-		return ctrl.Result{}, nil
+	created := v1.GetCondition(updatedQuay.Status.Conditions, v1.ConditionComponentsCreated)
+	migrationInProgress := created != nil && created.Reason == v1.ConditionReasonMigrationsInProgress
+	if migrationInProgress {
+		log.Info("++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++")
+		log.Info("migration currently in progress!")
+		log.Info(fmt.Sprintf("currentVersion is: '%s'", updatedQuay.Status.CurrentVersion))
+
+		err := wait.Poll(upgradePollInterval, upgradePollTimeout, func() (bool, error) {
```
ricardomaraschini: I don't think we need this `Poll`. We can check only once and if the migration is running return a reschedule, we will check again during the next cycle.

controllers/quay/quayregistry_controller.go
```diff
@@ -345,68 +419,8 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-					qcopy.Status.Conditions = v1.SetCondition(qcopy.Status.Conditions, condition)
-					qcopy.Status.CurrentVersion = v1.QuayVersionCurrent
-					r.EventRecorder.Event(qcopy, corev1.EventTypeNormal, string(v1.ConditionReasonHealthChecksPassing), msg)
-
-					if err = r.Client.Status().Update(ctx, qcopy); err != nil {
-						log.Error(err, "could not update QuayRegistry status with current version")
-						return false, nil
-					}
-
-					qcopy.Spec.Components = v1.EnsureComponents(qcopy.Spec.Components)
```
ricardomaraschini, line 399: I have seen that you removed this "function" call, this is what was triggering the new event. On this case I guess that we gonna need to, once the migration is finished, return a reschedule as well.

ricardomaraschini (changes requested):
This is a step in the right direction. I am not sure about the "MigrationsFailed" condition though (what happens during the next reconcile loop after this condition has been applied?).

flavianmissi:
> I am not sure about the "MigrationsFailed" condition though (what happens during the next reconcile loop after this condition has been applied?)

Yeah that's totally a blind spot in the current implementation. I think it would make sense to recreate the migration job if it has failed, but right now this is already handled by CurrentVersion / QUAY_VERSION... so we could probably remove it.
But from an end-user perspective it could be helpful: whenever the job fails the condition would pop up in the CR making it clear exactly what's up.

What do you think?

ricardomaraschini:
> > I am not sure about the "MigrationsFailed" condition though (what happens during the next reconcile loop after this condition has been applied?)
> 
> Yeah that's totally a blind spot in the current implementation. I think it would make sense to recreate the migration job if it has failed, but right now this is already handled by CurrentVersion / QUAY_VERSION... so we could probably remove it. But from an end-user perspective it could be helpful: whenever the job fails the condition would pop up in the CR making it clear exactly what's up.
> 
> What do you think?

I would say that as long as the Reconcile runs entirely in case of "MigrationsFailed" we are good. That would dispatch another job and the migration would run again.

pkg/kustomize/kustomize.go
flavianmissi: Hmm yes, true!
Additionally, `quay.Status.CurrentVersion != v1.QuayVersionCurrent && !migrationInProgress` will evaluate to true if the migration job has failed, since we change the condition reason to migration failed.
Given that we set `spec.restartPolicy: OnFailure` in the upgrade job, I don't think we ever want to recreate the job except when there are new migrations to run.
We should only recreate the upgrade job (to force the job to run again) when the following is true:
1. `QUAY_VERSION` differs from `status.currentVersion`
2. the current job has run successfuly, so `status.succeeded == 1`
What you think?

controllers/quay/quayregistry_controller.go
flavianmissi: good catch, will fix

controllers/quay/quayregistry_controller.go
flavianmissi: Good point, will fix.

controllers/quay/quayregistry_controller.go
flavianmissi: true!

pkg/kustomize/kustomize.go
flavianmissi: Looks like a made a wrong assumption here. The upgrade job overlay will not recreate the job, it'll just set quay-app replicas to 0.
I think this code can stay as it was.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			return ctrl.Result{RequeueAfter: time.Second * 1}, nil
+		}
+
+		// when the job fails, the next reconciliation to run will think
+		// that migrations are currently not running, since we change
+		// the condition reason to v1.ConditionReasonMigrationsFailed.
+		// this doesn't necessarily describe reality, because kube will
+		// retry failed jobs for us, but it's desired behaviour because
+		// when the migration job fails due to misconfiguration, then the
+		// reconcile function should be allowed to proceed.
```
flavianmissi, line 197: Adding to the comment - we're counting that the version check on [line 425](https://github.com/quay/quay-operator/blob/a94041d2b525518f5aa005e138eb7387a7019f94/controllers/quay/quayregistry_controller.go#L425-L431) will eventually be reached, setting the condition reason back to `v1.ConditionReasonMigrationsInProgress` and scheduling another reconcile. After the condition is set back to migration in progress we'll check the upgrade job again.

controllers/quay/quayregistry_controller.go
flavianmissi: done

controllers/quay/quayregistry_controller.go
flavianmissi: done

flavianmissi:
> I would say that as long as the Reconcile runs entirely in case of "MigrationsFailed" we are good. That would dispatch another job and the migration would run again.

This (the reconcile loop run again, in its entirety) is the desired behaviour. I have to find a way to test it though.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			ctx,
+			types.NamespacedName{
+				Name:      updatedQuay.GetName() + "-" + v1.QuayUpgradeJobName,
+				Namespace: updatedQuay.GetNamespace(),
+			},
+			&upgradeJob,
+		)
+		if err != nil {
+			log.Error(err, "could not retrieve Quay upgrade `Job`")
+			return ctrl.Result{RequeueAfter: time.Second * 1}, nil
```
ricardomaraschini: No need to multiply time.Second by 1. I am not going to comment everywhere but this happens in other parts of the PR.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				Type:               v1.ConditionComponentsCreated,
+				Status:             metav1.ConditionTrue,
+				Reason:             v1.ConditionReasonComponentsCreationSuccess,
+				Message:            msg,
+				LastUpdateTime:     metav1.Now(),
+				LastTransitionTime: metav1.Now(),
+			}
+			qcopy.Status.Conditions = v1.SetCondition(qcopy.Status.Conditions, condition)
+			qcopy.Status.CurrentVersion = v1.QuayVersionCurrent
+			r.EventRecorder.Event(qcopy, corev1.EventTypeNormal, string(v1.ConditionReasonHealthChecksPassing), msg)
```
ricardomaraschini: I think it is safe to remove this (if tests are depending on this then they should be fixed). What do you think?

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		// retry failed jobs for us, but it's desired behaviour because
+		// when the migration job fails due to misconfiguration, then the
+		// reconcile function should be allowed to proceed.
+		if upgradeJob.Status.Failed == 0 {
+			updatedQuay, err = r.updateWithCondition(
+				updatedQuay,
+				v1.ConditionComponentsCreated,
+				metav1.ConditionFalse,
+				v1.ConditionReasonMigrationsFailed,
+				"failed to run migrations")
```
ricardomaraschini: Nice to have: wouldn't it be cool if we could show up the Reason why it is failing? I think that job.Status.Conditions has a Failed condition with the Reason.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if err != nil {
+				log.Error(err, "failed to update `conditions` of `QuayRegistry`")
+			}
+			return ctrl.Result{RequeueAfter: time.Second * 1}, nil
+		}
+
+		log.Info("++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++")
```
ricardomaraschini: I think we can start removing these debug logs.

ricardomaraschini (changes requested):
A couple of comments otherwise we are looking very good in here. Remember to rename the PR to point to the right PROJQUAY card.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,91 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		var upgradeJob batchv1.Job
+		err := r.Client.Get(
+			ctx,
+			types.NamespacedName{
+				Name:      updatedQuay.GetName() + "-" + v1.QuayUpgradeJobName,
+				Namespace: updatedQuay.GetNamespace(),
+			},
+			&upgradeJob,
+		)
+		if err != nil {
```
ricardomaraschini, line 156: What if `errors.IsNotFound(err)` ? We might ended up with this stuck in "MigrationInProgress" if someone delete the job manually. I guess we could do something similar to what is done when the migration fails, some other status like you did for `ConditionReasonMigrationsFailed`

controllers/quay/quayregistry_controller.go
ricardomaraschini: Loved the comment!

controllers/quay/quayregistry_controller.go
flavianmissi: That's a great idea! Its got both a reason and a message. I chose to go with the message since we already have a reason (`ConditionReasonMigrationsFailed`).
From `batchv1.JobCondition` type:
```
// [...]
	// (brief) reason for the condition's last transition.
	// +optional
	Reason string `json:"reason,omitempty" protobuf:"bytes,5,opt,name=reason"`
	// Human readable message indicating details about last transition.
	// +optional
	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
}
```

controllers/quay/quayregistry_controller.go
flavianmissi: Good thinking!

controllers/quay/quayregistry_controller.go
flavianmissi: Yup, let's remove it and see where we end up.

jonathankingfc (dismissed):
LGTM

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,108 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+
+		if upgradeJob.Status.Active == 1 {
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if upgradeJob.Status.Succeeded == 1 {
+			log.Info("Quay upgrade complete, updating `status.currentVersion`")
+
+			var freshQuay v1.QuayRegistry
```
ricardomaraschini: I think we should keep using `updatedQuay` here, we don't need to get a fresh copy. We use `updatedQuay` down below (line 212) so we can remove this.

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,108 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if err != nil {
+			log.Error(err, "could not retrieve Quay upgrade `Job`")
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		if upgradeJob.Status.Active == 1 {
+			return ctrl.Result{RequeueAfter: time.Second}, nil
```
ricardomaraschini, line 163: It could be useful to log something here. We can see that the loop is being called but awaiting for the migration to finish. wdyt?

controllers/quay/quayregistry_controller.go
```diff
@@ -120,10 +120,108 @@ func (r *QuayRegistryReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				metav1.ConditionFalse,
+				v1.ConditionReasonMigrationsFailed,
+				msg)
+			if err != nil {
+				log.Error(err, "failed to update `conditions` of `QuayRegistry`")
+			}
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+
+		return ctrl.Result{RequeueAfter: time.Second}, nil
```
ricardomaraschini, line 219: We gonna hit here if `Active != 1 && Failed != 1 and Succeeded != 1`. I think that this should never happen so it could be useful to log something so we can see that we are under this condition. I am saying that because I really don't know if this condition could happen but better be safe than sorry. wdyt?

controllers/quay/quayregistry_controller.go
flavianmissi: good catch!

controllers/quay/quayregistry_controller.go
flavianmissi: yup, will add an info.

controllers/quay/quayregistry_controller.go
flavianmissi: agreed, we should never end up here. I'll come up with something to log.

ricardomaraschini (approved):

PR #595, opened 2021-12-07
chore: automate downstream manifest generation

flavianmissi:
Not very elegant, but it works.

```
make prepare-downstream-release RELEASE=3.6.2 REPLACES=3.6.1 \
QUAY_RELEASE=3.6.2 CLAIR_RELEASE=3.6.2 CHANNEL=stable-3.6
```

will create a cluster service version virtually identical to the one currently
located in bundle/downstream. ~the bundle/downstream directory has not
yet been deleted so that the CSV file there can serve as base for
comparing the generated one with the actual CSV used downstream.~
Use [this gist](https://gist.github.com/flavianmissi/67a6d3c89e969c89c11159fdf76f96ef) for a quick way to compare the bundle generated by the
command introduced in this PR with the downstream bundle.

to compare both CSVs:
```
vimdiff bundle/manifests/quay-operator.clusterserviceversion.yaml \
bundle/downstream/manifests/quay-operator.clusterserviceversion.yaml
```

note that we're also replacing `QUAY_DEFAULT_BRANDING` value from
'upstream' to 'redhat'. it seems the operator knows how to tell Quay
which is the correct logo to use. this is also done in the CSV itself,
and is still done because I'm actually not sure if the mechanism behind
QUAY_DEFAULT_BRANDING can be trusted.

Relates to PROJQUAY-2690


bundle/manifests/quay-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,285 @@
...
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.annotations['quay-version']
+                - name: QUAY_DEFAULT_BRANDING
+                  value: upstream
+                - name: RELATED_IMAGE_COMPONENT_QUAY
+                  value: quay.io/projectquay/quay:latest
+                - name: RELATED_IMAGE_COMPONENT_CLAIR
+                  value: quay.io/projectquay/clair:nightly
+                - name: RELATED_IMAGE_COMPONENT_BUILDER
```
flavianmissi, line 158: Builder and builder qemu components we previously _not_ in upstream versions of our CRD. I'm not sure why, so for simplicity sake I've added them.

ricardomaraschini (dismissed):

ricardomaraschini:
Ops, conflicts.

flavianmissi:
on it

ricardomaraschini (approved):

PR #607, opened 2021-12-21
chore: automate downstream manifest generation

flavianmissi:
make prepare-downstream-release RELEASE=3.6.2 REPLACES=3.6.1 QUAY_RELEASE=3.6.2 CLAIR_RELEASE=3.6.2 CHANNEL=stable-3.6

will create a cluster service version virtually identical to the one
located in bundle/downstream.

note that we're also replacing `QUAY_DEFAULT_BRANDING` value from
'upstream' to 'redhat'. it seems the operator knows how to tell Quay
which is the correct logo to use. this is also done in the CSV itself,
and is still done because I'm actually not sure if the mechanism behind
QUAY_DEFAULT_BRANDING can be trusted.


flavianmissi:
had to cherry pick manually due to conflict (redhat-3.6 CSV had changed, master CSV didn't)

ibazulic:
Will this allow us to choose the exact version to install or will this literary replace the old version with no way of choosing the exact z-stream?

flavianmissi:
> Will this allow us to choose the exact version to install or will this literary replace the old version with no way of choosing the exact z-stream?

This PR isn't really changing the operator version / update graph.

For downstream it will still to be possible to install previous z-streams of the operator. For upstream this isn't yet the case and needs to be fixed in the near future.

ricardomaraschini (approved):

PR #615, opened 2022-01-04
refactoring: making reconcile loop idempotent (PROJQUAY-2046)

ricardomaraschini:
Rescheduling the event whenever it is possible to do so. This assures
that the Reconcile loop will be called again.


ricardomaraschini:
Opened a new PR to address this: https://github.com/quay/quay-operator/pull/623

PR #619, opened 2022-01-05
bump: bumping k8s.io dependencies (PROJQUAY-2046)

ricardomaraschini:
In order to be able to create Jobs in a suspended state we need to bump all kubernetes dependencies so we can leverage this feature. This is an attempt to make possible, later on, to split the Job Creation and Job Dispatch into two distinct steps as an effort for the idempotent reconciler.

After bumping some code changes were necessary:

- use the right import URL for prometheus types
- pod "Probe" became "ProbeHandler"
- signingengine.SigningEngineFieldGroup{} does not exist anymore
- variables need to be removed (logic passed to middleware instead)
- krusty.MakeKustomizer() invocation has changed


pkg/middleware/middleware.go
```diff
@@ -110,6 +110,12 @@ func Process(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, obj cl
 		return obj, nil
 	}
 
+	if quayComponentLabel == "quay-builder-route" {
```
jonathankingfc: Is this related to a dependency upgrade? Or is this a hot fix for a pre-existing issue?

ricardomaraschini: this is a weird problem I found after upgrading kustomize. by some unknown reason these variables (understand as "kustomize variables" in the yaml files) could not be processed anymore by kustomize (it started to crash indicating that it could not find the source for the var value).

that being said: let me take another look at it and see if I can find out what is the root cause. I will get back to you.

ricardomaraschini: For the record, this is the error returned after the upgrade:
```
Error: field specified in var '{BUILDMAN_HOSTNAME ~G_v1_Service {metadata.annotations["quay-buildmanager-hostname"]}}' not found in corresponding resource
```

Seems like it can't locate `metadata.annotations["quay-buildmanager-hostname"]`

flavianmissi: Could the error be related to these changes?

* https://github.com/quay/quay-operator/pull/619/files#diff-5fa9a25597c790fbe319a456b2bfd3c7d8f0c0cd44085b40058f625b99482d6bL21-L27
* https://github.com/quay/quay-operator/pull/619/files#diff-aa2affe4e7dd819645bd9ffc3a800ba187ccc6f76fb97d5677560d981da90116L10

Or did you perhaps make the changes due to the error? 🤔

ricardomaraschini: @flavianmissi this is my fix, committed this morning. It turns out that in newer versions the syntax `a["b"]` is not supported, we need to use `a.b`. @jonathankingfc could you review it again?

flavianmissi: aha, good catch!

flavianmissi (approved):

jonathankingfc (approved):

pkg/middleware/middleware.go
jonathankingfc: LGTM

PR #623, opened 2022-01-08
refactoring: idempotent reconciler (PROJQUAY-2046)

ricardomaraschini:
Allowing Reconciler to run multiple times without generating side effects.
    
- Stop creating new secrets every time
- Stop running database upgrade jobs every time
- Stop creating a job to run database init job (enable extension)
- Stop panicking in some places (returning errors instead)
- Cleaning logging string (stop logging a ton of uneeded info)
- Small improvements everywhere
- Stop setting number of replicas on deployments
    - They were overridding whatever was being set by hpa
- Removed unused functions and constants


kustomize/base/quay.deployment.yaml
```diff
@@ -7,7 +7,6 @@ metadata:
   annotations:
     quay-component: base
 spec:
-  replicas: 2
```
ibazulic, line 10: This will once again change how the operator deploys Quay when HPA is disabled. We will once again be deploying 1 pod, while HPA enabled deployments deploy 2. We specifically changed this because customers asked for at least two pods every time when Quay is deployed via the operator.

controllers/quay/features.go
```diff
@@ -35,267 +33,311 @@ const (
...
+		Name:      fmt.Sprintf("%s-quay-datastore", quay.GetName()),
+		Namespace: quay.GetNamespace(),
+	}
 
-					return ctx, quay, err
-				}
+	var claims objectbucket.ObjectBucketClaimList
+	if err := r.Client.List(ctx, &claims); err != nil {
+		r.Log.Info("cluster does not support `ObjectBucketClaim` API")
+		return nil
```
flavianmissi: This feels like an error, but the message is logged as info, and no error is returned. What am I missing?

controllers/quay/features.go
```diff
@@ -35,267 +33,311 @@ const (
...
 
-				host := string(datastoreConfig.Data[datastoreBucketHostKey])
-				if strings.Contains(host, ".svc") && !strings.Contains(host, ".svc.cluster.local") {
-					r.Log.Info("`ObjectBucketClaim` is using in-cluster endpoint, ensuring we use the fully qualified domain name")
-					host = strings.ReplaceAll(host, ".svc", ".svc.cluster.local")
-				}
+		var datastoreSecret corev1.Secret
+		if err := r.Client.Get(ctx, dstorensn, &datastoreSecret); err != nil {
+			r.Log.Error(err, "unable to retrieve Quay datastore `Secret`")
+			return fmt.Errorf("awaiting for bucket claim to be processed")
```
flavianmissi, line 229: Should we also log this message on line 229? `"unable to retrieve Quay datastore Secret"` is quite cryptic on its own...

controllers/quay/features.go
```diff
@@ -35,267 +33,311 @@ const (
...
-				ctx.StorageBucketName = string(datastoreConfig.Data[datastoreBucketNameKey])
-				ctx.StorageHostname = host
-				ctx.StorageAccessKey = string(datastoreSecret.Data[datastoreAccessKey])
-				ctx.StorageSecretKey = string(datastoreSecret.Data[datastoreSecretKey])
-				ctx.ObjectStorageInitialized = true
-			}
+		var datastoreConfig corev1.ConfigMap
+		if err := r.Client.Get(ctx, dstorensn, &datastoreConfig); err != nil {
+			r.Log.Error(err, "unable to retrieve Quay datastore `ConfigMap`")
+			return fmt.Errorf("awaiting for bucket claim to be processed")
```
flavianmissi, line 235: ditto

controllers/quay/features.go
```diff
@@ -35,267 +33,311 @@ const (
...
 	}
 
-	return ctx, quay, nil
+	r.Log.Info("`ObjectBucketClaim` not found")
+	return nil
 }
 
-// TODO: Improve this once `builds` is a managed component.
-func (r *QuayRegistryReconciler) checkBuildManagerAvailable(ctx *quaycontext.QuayRegistryContext, quay *v1.QuayRegistry, configBundle map[string][]byte) (*quaycontext.QuayRegistryContext, *v1.QuayRegistry, error) {
+// checkRoutesAvailable verifies if the config bundle contains an etry pointing to the buildman
```
flavianmissi: s/checkRoutesAvailable/checkBuildManagerAvailable
s/etry/entry

controllers/quay/features.go
```diff
@@ -35,267 +33,311 @@ const (
...
 	var grafanaDashboardNamespace corev1.Namespace
-	if err := r.Client.Get(context.Background(), namespaceKey, &grafanaDashboardNamespace); err != nil {
-		msg := fmt.Sprintf("Unable to find the Grafana config namespace %s. Monitoring component disabled", GrafanaDashboardConfigNamespace)
+	if err := r.Client.Get(ctx, namespaceKey, &grafanaDashboardNamespace); err != nil {
+		msg := fmt.Sprintf(
+			"Unable to find the Grafana config namespace %s. "+
+				"Monitoring component disabled",
+			GrafanaDashboardConfigNamespace,
+		)
 		r.Log.Info(msg)
```
flavianmissi: Don't we already log this as an error on the caller's side?

flavianmissi (commented):
left some comments/questions, but overall it's looking good!

controllers/quay/features.go
ricardomaraschini: I kept this as is to maintain the behavior already in place. That being said I think it makes sense to return an error here (this will make the reconcile loop to fail and a proper status to be recorded). I will fix it.

controllers/quay/features.go
ricardomaraschini: I don't follow, what do you mean?

controllers/quay/features.go
ricardomaraschini: I don't follow, what do you mean?

controllers/quay/features.go
ricardomaraschini: done.

controllers/quay/features.go
ricardomaraschini: Removing the log.

controllers/quay/features.go
flavianmissi: When this happens, the logs will show an error saying "unable to retrieve Quay datastore `Secret`". That alone doesn't say much.
My suggestion would be to add something like the returned error's message to the log, i.e:
```
r.Log.Error(err, "unable to retrieve Quay datastore `Secret` - bucket claim not ready")
```
Or something in the likes.

ricardomaraschini: Fixed, PTAL.

controllers/quay/features.go
ricardomaraschini: Fixed, PTAL.

flavianmissi (dismissed):
LGTM

ricardomaraschini:
> LGTM

Thanks. I am going to wait for @jonathankingfc review as well as this is too big.

controllers/quay/features.go
```diff
@@ -35,267 +33,304 @@ const (
...
-		r.Log.Info("TLS cert/key pair not provided, will use default cluster wildcard cert")
+// checkManagedTLS verifies if provided bundle contains entries for ssl key and cert,
+// populate the data in provided QuayRegistryContext if found.
+func (r *QuayRegistryReconciler) checkManagedTLS(
+	qctx *quaycontext.QuayRegistryContext, bundle *corev1.Secret,
+) {
+	providedTLSCert := bundle.Data["ssl.cert"]
+	providedTLSKey := bundle.Data["ssl.key"]
+
+	if len(providedTLSCert) == 0 || len(providedTLSKey) == 0 {
```
jonathankingfc, line 91: Could these values potentially be nil? I think we can add the ok syntax to the map check in lines 88 and 89.

kustomize/base/quay.deployment.yaml
jonathankingfc, line 10: Is there a specific reason for lowering the amount of replicas?

kustomize/components/clair/clair.deployment.yaml
```diff
@@ -7,7 +7,6 @@ metadata:
     quay-component: clair
   name: clair-app
 spec:
-  replicas: 2
```
jonathankingfc, line 10: Same as above comment

jonathankingfc (changes requested):

controllers/quay/features.go
ricardomaraschini: I don't think this is needed. If indexes `ssl.cert` and `ssl.key` don't exist then `providedTLSCert` and `providedTLSKey` will be `nil` and `len()` of a `nil` slice is zero. Did I follow you reasoning or I missed something ?

ibazulic (changes requested):
This will once again change how the operator deploys Quay when HPA is disabled. We will once again be deploying 1 pod, while HPA enabled deployments deploy 2. We specifically changed this because customers asked for at least two pods every time when Quay is deployed via the operator.

kustomize/base/quay.deployment.yaml
ricardomaraschini: If not removed then we will have a concurrency problem with the HPA as we can't assume the number of replicas set by the HPA (assume for instance that the HPA set this field to `3`, we can't reset it back to `2`). i.e. if we don't set it keeps the value set by the HPA.

ricardomaraschini: I don't think that to have a fixed value here applies to all scenarios. Isn't it better to have this working through an override?

ibazulic: override through CRD would be neat :-)

ricardomaraschini: As agreed I added a new commit to this PR that resets the number of replicas to 2 if HPA is unmanaged. As soon as this one gets merged we can work in having the value of replicas set through an override in CRD. Can you resolve this one if you are OK with it?

https://github.com/quay/quay-operator/pull/623/commits/f9b0b28fc40975940c93d47d375c53ec2a507b90

ricardomaraschini: After discussing this with @ibazulic I have restored the old behavior with a small difference: only set the number of replicas to 2 if HPA is not managed. Can you resolve this if you are OK with it?

flavianmissi (approved):

jonathankingfc (approved):
LGTM

flavianmissi:
Do we want to backport this to `redhat-3.6` @ricardomaraschini ?

ricardomaraschini:
> Do we want to backport this to `redhat-3.6` @ricardomaraschini ?

@flavianmissi no. Better to keep it in 3.7 (too many changes to include in a minor). I initially thought it would be good to backport but due to the size I reconsidered.

PR #626, opened 2022-01-14
proxy: adding proxy support (PROJQUAY-1579)

ricardomaraschini:
Passing through proxy configuration received through OLM operator. This
PR passes HTTP_PROXY, HTTPS_PROXY and NO_PROXY from the operator env and
into operand env.

NO_PROXY, as passed by OLM, already contain rules to allow direct
connection to kubernetes internal services therefore this PR only adds a
rule allowing Clair to directly connect to Quay through its Route.

### DIFF
--- a/pkg/kustomize/kustomize.go
+++ b/pkg/kustomize/kustomize.go
@@ -14,6 +14,7 @@ import (
 	objectbucket "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1"
 	route "github.com/openshift/api/route/v1"
 	prometheusv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"golang.org/x/net/http/httpproxy"
 	apps "k8s.io/api/apps/v1"
 	autoscaling "k8s.io/api/autoscaling/v2beta2"
 	batchv1 "k8s.io/api/batch/v1"
@@ -37,6 +38,7 @@ import (
 const (
 	QuayRegistryNameLabel = "quay-operator/quayregistry"
 
+	proxyConfigPrefix                 = "quay-proxy-config"
 	configSecretPrefix                = "quay-config-secret"
 	registryHostnameAnnotation        = "quay-registry-hostname"
 	buildManagerHostnameAnnotation    = "quay-buildmanager-hostname"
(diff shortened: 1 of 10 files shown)


dmesser:
Can these environment variables still be overwritten at the `QuayRegistry` configuration, in case the Quay deployment should utilize another proxy than the cluster uses? I think `Subscription` would already allow overwriting environment variables but it would apply to all Quay instances deployed by this operator and the `Subscription` object might be out of reach for the individual managing `QuayRegistry`.

pkg/kustomize/kustomize.go
```diff
@@ -312,7 +314,25 @@ func KustomizationFor(
 		quayConfigTLSSources = append(quayConfigTLSSources, "ssl.key="+string(ctx.TLSKey))
 	}
 
+	// read the proxy configuration from the environment. makes sure we add the quay server
+	// host name to the list of addresses reachable without proxy (NO_PROXY).
+	proxyenv := httpproxy.FromEnvironment()
+	addrs := strings.Split(proxyenv.NoProxy, ",")
```
dmage: note that if NoProxy is "", then addrs will be `[]string{""}`, not `[]string{}`.

pkg/kustomize/kustomize.go
```diff
@@ -37,6 +38,7 @@ import (
 const (
 	QuayRegistryNameLabel = "quay-operator/quayregistry"
 
+	proxyConfigPrefix                 = "quay-proxy-config"
```
dmage, line 41: What will happen if this config map doesn't exist?

dmage: Nvmd, it is a generated secret.

pkg/kustomize/secrets.go
```diff
@@ -52,7 +52,9 @@ func FieldGroupFor(
 
 		fieldGroup.FeatureSecurityScanner = true
 		fieldGroup.SecurityScannerV4Endpoint = fmt.Sprintf(
-			"http://%s-clair-app:80", quay.GetName(),
+			"http://%s-clair-app.%s.svc.cluster.local",
```
flavianmissi, line 56: I realize you didn't introduce this, but is there a better way to get this address? It feels fragile.

flavianmissi (commented):
Looks good to me, but I think you want to address [this comment](https://github.com/quay/quay-operator/pull/626/files#r784755126) @ricardomaraschini

ricardomaraschini:
> Can these environment variables still be overwritten at the `QuayRegistry` configuration, in case the Quay deployment should utilize another proxy than the cluster uses? I think `Subscription` would already allow overwriting environment variables but it would apply to all Quay instances deployed by this operator and the `Subscription` object might be out of reach for the individual managing `QuayRegistry`.

This "override" feature is part of https://issues.redhat.com/browse/PROJQUAY-3055, also planned for v3.7.0 and the goal is to deliver both of them (proxy support + env override) together.

pkg/kustomize/secrets.go
ricardomaraschini: What do you suggest?

flavianmissi: Yeah that's the thing, I don't really know a better way either.

pkg/kustomize/kustomize.go
ricardomaraschini: Addressed. PTAL.

pkg/kustomize/secrets.go
ricardomaraschini: AFAIK the domain "cluster.local" always exist in a cluster, the host "svc" in the domain indicates that it is a service, then we have the namespace and the service name:

```
<servicename>.<namespace>.svc.cluster.local
```

Host "svc.cluster.local" is automatically added to the NO_PROXY list. @dmage do you know if this domain can be customized to something else?

flavianmissi (dismissed):

pkg/kustomize/secrets.go
ricardomaraschini: As discussed in our Operators meeting we can keep this hardcoded, therefore I am resolving this one.

/cc @dmage @flavianmissi

pkg/kustomize/kustomize.go
ricardomaraschini: Are you happy with the proposed solution? If yes, can you please "resolve" this one?

flavianmissi (approved):

PR #629, opened 2022-01-17
clair: use Lint instead of Validate

hdonnay:
Accidentally used the Validate function, which sets defaults, instead of
Lint, which simply reports warnings.


jonathankingfc (approved):
LGTM

PR #631, opened 2022-01-24
postgres: disable huge pages (PROJQUAY-2869)

ricardomaraschini:
As per https://bugzilla.redhat.com/show_bug.cgi?id=2032811 we can't
deploy postgres in a cluster whose nodes have huge pages enabled.

This PR disables huge pages support in postgres, this is a work around
and can be removed in the future (as soon as the underlying issue is
fixed in the kernel/kubelet).


flavianmissi (approved):
LGTM

PR #633, opened 2022-01-24
services: patch component services (PROJQUAY-3083)

ricardomaraschini:
Start treating services as regular objects. We were treating them as
immutable objects, therefore we were deleting and creating them every
time. With this patch we start to call Patch() for service objects as
we do for all regular objects.


flavianmissi (approved):
Nice catch.

PR #635, opened 2022-01-25
env: support env var overrides (PROJQUAY-3055)

ricardomaraschini:
Overrides env variables in deployments. Each component has its own env
override property. Base component overrides live in the root of the
spec.


flavianmissi (commented):
Is there a point to try and automate some basic tests for this feature?
Looking good otherwise.

apis/quay/v1/quayregistry_types.go
```diff
@@ -83,10 +83,13 @@ const (
...
 // QuayRegistrySpec defines the desired state of QuayRegistry.
 type QuayRegistrySpec struct {
-	// ConfigBundleSecret is the name of the Kubernetes `Secret` in the same namespace which contains the base Quay config and extra certs.
+	// ConfigBundleSecret is the name of the Kubernetes `Secret` in the same namespace
+	// which contains the base Quay config and extra certs.
 	ConfigBundleSecret string `json:"configBundleSecret,omitempty"`
 	// Components declare how the Operator should handle backing Quay services.
 	Components []Component `json:"components,omitempty"`
+	// Overrides holds overrides for the Base component (quay-app).
+	Overrides *Override `json:"overrides,omitempty"`
```
jonathankingfc: Not sure why this line is showing as an addition. You might have to rebase?

jonathankingfc (changes requested):
I would also add validation to the overrides to make sure that the specific component supports env overrides. For example, the route and objectstorage component will not support 'env' overrides. See: https://github.com/quay/quay-operator/blob/12dc87b91272adb8cff9f21090cb0d5fa55f448a/apis/quay/v1/quayregistry_types.go#L381

apis/quay/v1/quayregistry_types.go
ricardomaraschini: This `Override` is for the `base` component (there is no such thing as `base` in the list of components so I put it in the spec's root). I am not happy with it but I could not think in any other approach (to add a `base` component to `spec.components` I think is a no go as it can't be "unmanaged"). Do you have a better idea?

ricardomaraschini:
> I would also add validation to the overrides to make sure that the specific component supports env overrides. For example, the route and objectstorage component will not support 'env' overrides. See:
> 
> https://github.com/quay/quay-operator/blob/12dc87b91272adb8cff9f21090cb0d5fa55f448a/apis/quay/v1/quayregistry_types.go#L381

Good stuff @jonathankingfc. I have added the validation logic, PTAL.

apis/quay/v1/quayregistry_types.go
jonathankingfc: I actually prefer the approach of a `base` component. It fits more nicely with the current API even if it can't be set to unmanaged. We have an idea of what a `base` component would look like as its used internally by the operator.

jonathankingfc (changes requested):
Can you add an e2e test to this as well? You can do an assertion on the deployment

apis/quay/v1/quayregistry_types.go
dmage: `base` = `quay-app`? Why can't it be unmanaged?

ricardomaraschini: It can't be "unmanaged" without creating a ton of problems for us at this stage (mirror will crash is one example). I have created a component "base" but forced it to be managed (rolls back if user sets it to "unmanaged"). To support such a component as "unmanaged" is a totally different task and I won't be addressing it here.

ricardomaraschini:
> Can you add an e2e test to this as well? You can do an assertion on the deployment

We absolutely can! I added a couple, could you check them?

ricardomaraschini:
> Is there a point to try and automate some basic tests for this feature? Looking good otherwise.

Took me a while but I added a couple of e2e tests.

apis/quay/v1/quayregistry_types.go
ricardomaraschini: I will be opening a new Jira Card to deal with this "unmanaged base component" thing. Resolving this conversation so we can get this in.

jonathankingfc (approved):
LGTM

PR #636, opened 2022-01-25
config: unset config endpoint/unmanaged route (PROJQUAY-2049)

ricardomaraschini:
Only sets status.configEditorEndpoint if cluster supports Routes and the
Route component is managed.


flavianmissi (approved):
LGTM.
Would be nice with some unit tests tho 😄

PR #637, opened 2022-01-25
routes: improve error reporting (PROJQUAY-2067)

ricardomaraschini:
Adding a proper error message when unable to parse config.yaml during
Route detection.


flavianmissi (approved):
Beautiful. Are we backporting it to 3.6? Seems trivial enough to me, given there aren't any conflicts.

PR #642, opened 2022-02-10
chore: stop running operator during e2e (PROJQUAY-3055)

ricardomaraschini:
Stop running the operator when we run the e2e tests. These are going to
run agains a cluster where the operator is already installed.


ricardomaraschini:
Seems like we are not running e2e tests for the nightly but for every PR instead. Closing this one as it is not right.

PR #643, opened 2022-02-11
Removed e2e from pull_request

ricardomaraschini:
We need to address this somehow but for now let's keep this as a nightly
job only thing.


ricardomaraschini:
@flavianmissi would you mind checking if this is right ?

flavianmissi (dismissed):
Is the intention to run the `make test-e2e` on the e2e nightly workflow on a follow-up PR?
If that's the case, then this one LGTM.

ricardomaraschini:
> Is the intention to run the `make test-e2e` on the e2e nightly workflow on a follow-up PR? If that's the case, then this one LGTM.

@flavianmissi I thought it was already enabled for nightly builds. Added, PTAL.

flavianmissi (approved):
🚀

PR #644, opened 2022-02-14
Do not run e2e on deploy.sh

ricardomaraschini:
Deploy script should worry about deploying the quay operator and not
deploying and running the E2E tests.


flavianmissi (dismissed):
nice catch!

flavianmissi (approved):

PR #649, opened 2022-02-17
reconcile: immutable resources recreation (PROJQUAY-3093)

ricardomaraschini:
When deleting an immutable resource we don't need to check if the error
returned is of type AlreadyExists. This commit also adds some logging so
it is clear what is happening during immutable resources recreation.


ricardomaraschini:
I am working in a new version for this PR

PR #657, opened 2022-02-23
scale: allow users to scale down components (PROJQUAY-3056)

ricardomaraschini:
Allows users to scale down the following components at any time:

- Clair
- Base (clair)
- Mirror

Allows users to scale up any of these as well if HPA component is
unmanaged.


jonathankingfc (approved):
LGTM,approved

PR #662, opened 2022-03-04
monitoring: improved coverage (PROJQUAY-3290)

ricardomaraschini:
Improved the coverage for Deployment health checks. We now take into
account possible overrides.

### DIFF
--- a/pkg/cmpstatus/clair.go
+++ b/pkg/cmpstatus/clair.go
@@ -73,6 +73,30 @@ func (c *Clair) Check(ctx context.Context, reg qv1.QuayRegistry) (qv1.Condition,
 		}, nil
 	}
 
+	// users are able to override the number of replicas. if they do override it to zero
+	// we expect zero replicas to be running.
+	replicas := qv1.GetReplicasOverrideForComponent(&reg, qv1.ComponentClair)
+	scaleddown := replicas != nil && *replicas == 0
+	if scaleddown {
+		if dep.Status.AvailableReplicas == 0 {
+			return qv1.Condition{
+				Type:           qv1.ComponentClairReady,
+				Reason:         qv1.ConditionReasonComponentReady,
+				Status:         metav1.ConditionTrue,
+				Message:        "Clair manually scaled down",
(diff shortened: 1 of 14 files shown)


pkg/cmpstatus/clair.go
```diff
@@ -73,6 +73,30 @@ func (c *Clair) Check(ctx context.Context, reg qv1.QuayRegistry) (qv1.Condition,
...
+	// users are able to override the number of replicas. if they do override it to zero
+	// we expect zero replicas to be running.
+	replicas := qv1.GetReplicasOverrideForComponent(&reg, qv1.ComponentClair)
+	scaleddown := replicas != nil && *replicas == 0
+	if scaleddown {
+		if dep.Status.AvailableReplicas == 0 {
+			return qv1.Condition{
+				Type:           qv1.ComponentClairReady,
+				Reason:         qv1.ConditionReasonComponentReady,
+				Status:         metav1.ConditionTrue,
```
dmage, line 85: Base reports itself as Ready: false when it's scaled down, but Clair as Ready: true. Is it intentionally?

pkg/cmpstatus/postgres.go
```diff
@@ -41,6 +41,11 @@ func (p *Postgres) Check(ctx context.Context, reg qv1.QuayRegistry) (qv1.Conditi
 	}
 
 	nsn := types.NamespacedName{
+		Namespace: reg.Namespace,
+		Name:      fmt.Sprintf("%s-quay-app-upgrade", reg.Name),
+	}
+
+	nsn = types.NamespacedName{
```
dmage: What's going on here?..

pkg/cmpstatus/clair.go
dmage: nvmd, I misread changes in pkg/cmpstatus/base.go.

pkg/cmpstatus/postgres.go
ricardomaraschini: I don't know, I don't know! Fixed.

dmage (approved):

PR #663, opened 2022-03-06
state: use the same secret to store managed keys (PROJQUAY-3358)

ricardomaraschini:
This PR stops appending the hash to the "managed keys" secret name. By
doing so we stop creating a new secret every time a reconcile runs and
we use always the same.


jonathankingfc (approved):
LGTM

PR #666, opened 2022-03-14
routes: fail only when managed (PROJQUAY-3399)

ricardomaraschini:
On v3.6.x we only fail when checking if routes are available if the
route component is managed. This behavior has been lost during the
migration to v3.7.0. This commit restores the old behavior.


ricardomaraschini:
I had a problem with a config.yaml so this PR is not valid

PR #668, opened 2022-03-16
tests: e2e test for HPA component (PROJQUAY-3408)

nichitagutu:
added e2e test for HPA component

### DIFF
--- a/e2e/hpa/00-assert.yaml
+++ b/e2e/hpa/00-assert.yaml
@@ -1,0 +1,93 @@
+apiVersion: quay.redhat.com/v1
+kind: QuayRegistry
+metadata:
+  finalizers:
+  - quay-operator/finalizer
+  name: hpa
+spec:
+  components:
+  - kind: clair
+    managed: true
+  - kind: clairpostgres
+    managed: true
+  - kind: quay
+    managed: true
+  - kind: postgres
+    managed: true
+  - kind: redis
+    managed: true
+  - kind: horizontalpodautoscaler
+    managed: true
+  - kind: objectstorage
+    managed: true
+  - kind: route
+    managed: true
+  - kind: mirror
+    managed: true
+  - kind: monitoring
+    managed: true
+  - kind: tls
+    managed: true
+status:
+  conditions:
+  - message: Horizontal pod autoscaler found
+    reason: ComponentReady
+    status: "True"
+    type: ComponentHPAReady
+  - message: Route admitted
+    reason: ComponentReady
+    status: "True"
+    type: ComponentRouteReady
+  - message: ServiceMonitor and PrometheusRules created
+    reason: ComponentReady
+    status: "True"
+    type: ComponentMonitoringReady
+  - message: Deployment hpa-quay-database healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentPostgresReady
+  - message: Object bucket claim bound
+    reason: ComponentReady
+    status: "True"
+    type: ComponentObjectStorageReady
+  - message: Clair component healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentClairReady
+  - message: ClairPostgres component healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentClairPostgresReady
+  - message: Using cluster wildcard certs
+    reason: ComponentReady
+    status: "True"
+    type: ComponentTLSReady
+  - message: Deployment hpa-quay-redis healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentRedisReady
+  - message: Quay component healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentBaseReady
+  - message: Deployment hpa-quay-mirror healthy
+    reason: ComponentReady
+    status: "True"
+    type: ComponentMirrorReady
+  - message: All components reporting as healthy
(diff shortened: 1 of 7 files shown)


e2e/hpa/01-unmanage-hpa.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - kind: route
+    managed: true
+  - kind: mirror
+    managed: true
+  - kind: monitoring
+    managed: true
+  - kind: tls
+    managed: true
+---
+apiVersion: autoscaling/v1
```
ricardomaraschini, line 31: Why do we create this HPA here ?

e2e/hpa/00-assert.yaml
```diff
@@ -0,0 +1,88 @@
...
+    status: "True"
+    type: Available
+  - message: All registry components created
+    reason: ComponentsCreationSuccess
+    status: "True"
+    type: ComponentsCreated
+  - message: All objects created/updated successfully
+    reason: ComponentsCreationSuccess
+    status: "False"
+    type: RolloutBlocked
```
ricardomaraschini, line 88: I think that this assert should contain the HPA as well. The HPA creation is expected, no?

e2e/hpa/01-assert.yaml
```diff
@@ -0,0 +1,98 @@
...
+    type: ComponentsCreated
+  - message: All objects created/updated successfully
+    reason: ComponentsCreationSuccess
+    status: "False"
+    type: RolloutBlocked
+---
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: hpa-quay-mirror
```
ricardomaraschini, line 110: This does not seem right. When the QuayRegistry is created we expect to see three HPAs and their names are:

```
X-clair-app
X-quay-app
X-quay-mirror
```
Where X is equal to QuayRegistry.Spec.Name.

e2e/hpa/02-assert.yaml
```diff
@@ -0,0 +1,101 @@
...
+kind: HorizontalPodAutoscaler
+metadata:
+  annotations:
+    quay-component: mirror
+  labels:
+    quay-operator/quayregistry: hpa
+status:
+  currentReplicas: 2
+  desiredReplicas: 2
+---
```
ricardomaraschini, line 114: This one seems to be wrong as well. We do not specify any name for the HPA so we don't know what HPA is used to assert.

ricardomaraschini (changes requested):

e2e/hpa/00-assert.yaml
nichitagutu: Doesn't the first message assume that the HPA is created?

`message: Horizontal pod autoscaler found`

> I think that this assert should contain the HPA as well. The HPA creation is expected, no?

ricardomaraschini: Yep, that makes sense.

e2e/hpa/01-assert.yaml
nichitagutu: I've thought that testing against 1 hpa would be enough, would write all 3 of them then.

> This does not seem right. When the QuayRegistry is created we expect to see three HPAs and their names are:
> 
> ```
> X-clair-app
> X-quay-app
> X-quay-mirror
> ```
> 
> Where X is equal to QuayRegistry.Spec.Name.

e2e/hpa/02-assert.yaml
nichitagutu: True, name is not specified, but doesn't annotations and labels specify which hpa the test asserts (hpa-quay-mirror in this case) or there could be more than one hpa per component?
```
  annotations:
    quay-component: mirror
  labels:
    quay-operator/quayregistry: hpa

File: e2e/hpa/01-unmanage-hpa.yaml (hunk 1 above)
Comment (nichitagutu) [reply e2e/hpa/01-unmanage-hpa.yaml:31]: > Why do we create this HPA here ?

I'm not sure what creation you're mentioning.

```
apiVersion: autoscaling/v1
kind: HorizontalPodAutoscaler
metadata:
  name: hpa-quay-mirror
spec:
  minReplicas: 1
  ```
  If you mean this, these lines update the existing hpa, existence of which is asserted in the first step.

ricardomaraschini:
@Nikita757 https://github.com/quay/quay-operator/pull/671 was merged so you gonna need to update references to "base" component to "quay" for this one to get in. Let me know when I can give this one a final review.

nichitagutu:
@ricardomaraschini Updated the references

e2e/hpa/00-assert.yaml
```diff
@@ -0,0 +1,93 @@
...
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
```
ricardomaraschini, line 91: Why are we adding this TestAssert here?

ricardomaraschini (commented):
This looks awesome! There is one final question then we can merge this. See my comment.

nichitagutu:
> This looks awesome! There is one final question then we can merge this. See my comment.
> Why are we adding this TestAssert here?

I've added it to increase timeout for this particular test, because otherwise 500 seconds may not be enough for everything to get up and running.

ricardomaraschini (approved):

PR #671, opened 2022-03-23
component: renamine base to quay (PROJQUAY-3458) 

ricardomaraschini:
This commit renames the Base component into Quay component.


HammerMeetNail (approved):
LGTM 👍

PR #673, opened 2022-03-29
upgradejob: report job status (PROJQUAY-3443)

ricardomaraschini:
Take the upgrade job status into account when assessing quay component
status. This will show up quay in a non ready state until the database
upgrade job has been finished successfully.


jonathankingfc (approved):
Nice stuff, approved!

PR #675, opened 2022-03-30
clairpostgres: fail if no config is provided (PROJQUAY-3441)

ricardomaraschini:
If clairpostgres component is unmanaged, clair is managed but no clair
database information has been provided by the user we now fail.

### DIFF
--- a/apis/quay/v1/quayregistry_types.go
+++ b/apis/quay/v1/quayregistry_types.go
@@ -30,12 +30,18 @@ import (
 	quaycontext "github.com/quay/quay-operator/pkg/context"
 )
 
+// QuayVersion represents a quay version as a string. Normally this is set using semantic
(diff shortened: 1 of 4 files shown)


apis/quay/v1/quayregistry_types.go
```diff
@@ -30,12 +30,18 @@ import (
 	quaycontext "github.com/quay/quay-operator/pkg/context"
 )
 
+// QuayVersion represents a quay version as a string. Normally this is set using semantic
+// versioning, e.g. 3.7.0.
```
dmage, line 34: is it a build-time constant?

ricardomaraschini: nope, the value is set at runtime (env var)

jonathankingfc (approved):
LGTM

PR #680, opened 2022-03-31
chore: Update controller-gen from v0.5.0 to v0.8.0 (PROJQUAY-3068)

jonathankingfc:
- Update controller-gen from v0.5.0 to v0.8.0
- Add rule to makefile to track crd in bundle directory

### DIFF
--- a/bundle/manifests/quayregistries.crd.yaml
+++ b/bundle/manifests/quayregistries.crd.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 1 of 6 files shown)


bundle/manifests/quayregistries.crd.yaml
```diff
@@ -1,8 +1,9 @@
+---
```
ricardomaraschini: @flavianmissi isnt' this going to break the whole installation? @jonathankingfc I think we need to remove this line from all files, I remember that OLM can't find the CRD if it contains this thingy here.

ricardomaraschini (changes requested):

ricardomaraschini (approved):

PR #685, opened 2022-04-11
clair: fix clair config issue (PROJQUAY-3430)

ricardomaraschini:
Clair structs can't even serialize and unserialize themselves properly.
If we use its exposed Config struct it uses LogLevel as int but the
application (Clair) needs it to be a string.

### DIFF
--- a/pkg/kustomize/secrets.go
+++ b/pkg/kustomize/secrets.go
@@ -385,38 +385,38 @@ func clairConfigFor(log logr.Logger, quay *v1.QuayRegistry, quayHostname, preSha
 		return nil, err
 	}
 
-	cfg := config.Config{
-		HTTPListenAddr: ":8080",
-		LogLevel:       config.InfoLog,
-		Indexer: config.Indexer{
-			ConnString:           dbConn,
-			ScanLockRetry:        10,
-			LayerScanConcurrency: 5,
-			Migrations:           true,
+	cfg := map[string]interface{}{
+		"http_listen_addr": ":8080",
+		"log_level":        "info",
+		"indexer": map[string]interface{}{
+			"connstring":             dbConn,
+			"scanlock_retry":         10,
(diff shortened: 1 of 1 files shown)


jonathankingfc:
LGTM, but I am wondering if we need to drop the clair config struct completely? @hdonnay Would this function be called during a yaml unmarshal? https://github.com/quay/clair/blob/main/config/enums.go#L60 Or do you think there is a better way of generating a clair config from this module?

hdonnay:
`Gopkg.in/yaml.v2` and `gopkg.in/yaml.v3` both use the `TextMarshaler` interface, so it should round-trip the same way it does with json, e.g. 

```go
package config_test

import (
	"encoding/json"
	"fmt"
	"strings"

	"github.com/quay/clair/config"
)

func ExampleMarshal() {
	var cfg config.Config
	var b strings.Builder

	if err := json.NewEncoder(&b).Encode(&cfg); err != nil {
		panic(err)
	}
	fmt.Println(b.String())
	// Output:
	// {"http_listen_addr":"","introspection_addr":"","log_level":"info","indexer":{"scanner":{},"connstring":""},"matcher":{"connstring":"","indexer_addr":"","update_retention":0},"matchers":{"names":null},"updaters":{"filter":""},"notifier":{"connstring":"","indexer_addr":"","matcher_addr":""},"auth":{},"trace":{"name":"","jaeger":{"agent":{"endpoint":""},"collector":{"endpoint":""}}},"metrics":{"prometheus":{},"name":""}}
}
```

That said, the zero value of a `config.Config` isn't the defaults, so regardless of whether you populate it and render it or do something else, the code should probably still call `Lint`.

ricardomaraschini:
@jonathankingfc we gain absolutely nothing by bringing that struct back.

jonathankingfc (approved):
lgtm

PR #694, opened 2022-04-20
overrides: Allow labels and annotations overrides (PROJQUAY-2897)

nichitagutu:
Allows changing labels and annotations on QuayRegistry.

https://issues.redhat.com/browse/PROJQUAY-2897


apis/quay/v1/quayregistry_types.go
```diff
@@ -126,9 +126,11 @@ type Component struct {
...
 // Override describes configuration overrides for the given managed component
 type Override struct {
-	VolumeSize *resource.Quantity `json:"volumeSize,omitempty"`
-	Env        []corev1.EnvVar    `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
-	Replicas   *int32             `json:"replicas,omitempty"`
+	VolumeSize  *resource.Quantity `json:"volumeSize,omitempty"`
+	Env         []corev1.EnvVar    `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
+	Replicas    *int32             `json:"replicas,omitempty"`
+	Labels      map[string]string  `json:"labels,omitempty" protobuf:"bytes,11,rep,name=labels"`
+	Annotations map[string]string  `json:"annotations,omitempty" protobuf:"bytes,12,rep,name=annotations"`
```
ricardomaraschini: You can remove the `protobuf` annotations here, they are not needed.

apis/quay/v1/quayregistry_types.go
```diff
@@ -752,6 +754,36 @@ func GetEnvOverrideForComponent(quay *QuayRegistry, kind ComponentKind) []corev1
 	return nil
 }
 
+func GetLabelsOverrideForComponent(quay *QuayRegistry, kind ComponentKind) map[string]string {
```
ricardomaraschini, line 759: Haven't checked but I believe this is generating a warning on my lint here. We should always add comments to "exported" functions (the ones that start with an uppercase). I would suggest you to read the [Effective Go](https://go.dev/doc/effective_go), for this specific case the section "Commentary".

apis/quay/v1/quayregistry_types.go
```diff
@@ -752,6 +754,36 @@ func GetEnvOverrideForComponent(quay *QuayRegistry, kind ComponentKind) []corev1
...
+		if cmp.Overrides == nil {
+			return nil
+		}
+
+		return cmp.Overrides.Labels
+	}
+	return nil
+}
+
+func GetAnnotationsOverrideForComponent(quay *QuayRegistry, kind ComponentKind) map[string]string {
```
ricardomaraschini, line 788: Same as my previous comment regarding comments. AH AH.

pkg/middleware/middleware.go
```diff
@@ -125,6 +125,14 @@ func Process(quay *v1.QuayRegistry, obj client.Object, skipres bool) (client.Obj
 			return nil, err
 		}
 
+		if olabels := v1.GetLabelsOverrideForComponent(quay, kind); olabels != nil {
+			dep.Labels = olabels
```
ricardomaraschini: We should not fully override this property. I think we should do something like this:

```golang
for key, val := range olabels {
    dep.Labels[key] = val
}
```

Also, make sure `dep.Label` is not nil prior to this otherwise it will `panic`.

pkg/middleware/middleware.go
```diff
@@ -125,6 +125,14 @@ func Process(quay *v1.QuayRegistry, obj client.Object, skipres bool) (client.Obj
...
+		}
+
+		if oannot := v1.GetAnnotationsOverrideForComponent(quay, kind); oannot != nil {
+			dep.Annotations = oannot
+		}
+
```
ricardomaraschini, line 162: Same as my previous [comment](https://github.com/quay/quay-operator/pull/694/files#diff-5b6d6847c385671f06ecd44295ce5691596b68f54f3aeee404691bb09779be1eR129)

ricardomaraschini (changes requested):
Right on track Nikita! Thanks for that. I have left a few comments for you to check.

pkg/middleware/middleware.go
```diff
@@ -125,6 +125,26 @@ func Process(quay *v1.QuayRegistry, obj client.Object, skipres bool) (client.Obj
...
+			if dep.Labels == nil {
+				dep.Labels = olabels
+			} else {
+				for key, value := range olabels {
+					dep.Labels[key] = value
+				}
+			}
+		}
+
+		if oannot := v1.GetAnnotationsOverrideForComponent(quay, kind); oannot != nil {
```
ricardomaraschini, line 154:
```golang
if dep.Annotations == nil {
    dep.Annotations := map[string]string{}
}
```

pkg/middleware/middleware.go
```diff
@@ -140,6 +140,24 @@ func Process(quay *v1.QuayRegistry, obj client.Object, skipres bool) (client.Obj
 			return nil, err
 		}
 
+		if olabels := v1.GetLabelsOverrideForComponent(quay, kind); olabels != nil {
+			if dep.Labels == nil {
+				dep.Labels = map[string]string{}
+			}
+			for key, value := range olabels {
+				dep.Labels[key] = value
```
ricardomaraschini: I think that if we allow any override we can have other problems. I would say that the labels "quay-component", "app" and "quay-operator/quayregistry" should never be overwritten by the user. Can we get these exception here ?

apis/quay/v1/quayregistry_types.go
```diff
@@ -752,6 +754,40 @@ func GetEnvOverrideForComponent(quay *QuayRegistry, kind ComponentKind) []corev1
 	return nil
 }
 
+// GetLabelsOverrideForComponent returns overriden labels for the provided component
+// nil is returned if there are no label overrides
+func GetLabelsOverrideForComponent(quay *QuayRegistry, kind ComponentKind) map[string]string {
```
ricardomaraschini, line 759: The sad story here is that as the config-tool is also part of the "quay" component the same labels will apply for both of them :-(

ricardomaraschini (changes requested):

pkg/middleware/middleware.go
nichitagutu: Should be handled now

ricardomaraschini:
Replaced by https://github.com/quay/quay-operator/pull/726

PR #695, opened 2022-04-20
tls: Create secret for shared ca certs (PROJQUAY-3593)

jonathankingfc:
- Any key in the provided configBundleSecret that starts with the prefix 'extra_ca_cert_*' will be put in a shared secret called extra-ca-certs
- This will be mounted inside the pods and enables proxy certs to be shared

### DIFF
--- a/kustomize/components/job/quay.upgrade.job.yaml
+++ b/kustomize/components/job/quay.upgrade.job.yaml
@@ -20,8 +20,14 @@ spec:
           secret:
             secretName: quay-config-secret
         - name: extra-ca-certs
-          configMap:
-            name: cluster-service-ca
+          projected:
+            sources:
+              - configMap:
+                  name: cluster-service-ca
+              - configMap:
+                  name: cluster-trusted-ca
+              - secret:
+                  name: extra-ca-certs  
       containers:
         - name: quay-app-upgrade
           image: quay.io/projectquay/quay@sha256:5660d7174218e1cb21bf6ef406602dbe8c01c878c630a9f310fe3e5560d4c2cd
--- a/kustomize/components/mirror/mirror.deployment.yaml
+++ b/kustomize/components/mirror/mirror.deployment.yaml
@@ -33,6 +33,8 @@ spec:
                   name: cluster-service-ca
               - configMap:
                   name: cluster-trusted-ca
+              - secret:
+                  name: extra-ca-certs      
       initContainers:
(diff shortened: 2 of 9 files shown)


kustomize/components/job/quay.upgrade.job.yaml
```diff
@@ -20,8 +20,15 @@ spec:
...
-            name: cluster-service-ca
+          projected:
+            sources:
+              - configMap:
+                  name: cluster-service-ca
+              - configMap:
+                  name: cluster-trusted-ca
+              - secret:
+                  name: extra-ca-certs
+
```
ricardomaraschini: Could you remove this empty line ?

kustomize/components/mirror/mirror.deployment.yaml
```diff
@@ -33,6 +33,9 @@ spec:
                   name: cluster-service-ca
               - configMap:
                   name: cluster-trusted-ca
+              - secret:
+                  name: extra-ca-certs
+
```
ricardomaraschini: Remove the empty line ?

pkg/kustomize/secrets.go
```diff
@@ -352,10 +352,12 @@ func componentConfigFilesFor(log logr.Logger, qctx *quaycontext.QuayRegistryCont
...
-		for key, val := range configFiles {
-			if strings.HasPrefix(key, "clair_extra_ca_cert_") {
-				cfgFiles["clair_extra_ca_cert_"] = val
-			}
+		// Add ssl key and cert to bundle
+		if val, ok := configFiles["clair-ssl.crt"]; ok {
+			cfgFiles["clair-ssl.crt"] = val
+		}
+		if val, ok := configFiles["clair-ssl.key"]; ok {
+			cfgFiles["clair-ssl.key"] = val
```
ricardomaraschini, line 360: I don't think I understand these two `clair-ssl.{key,crt}` files. What are they used for ?

ricardomaraschini (changes requested):
This looks awesome! Thanks for implementing it, this whole cert management thing was driving me nuts. I left a few comments I would like to get addressed but it is already a darn good improvement over the mess we used to have.

pkg/kustomize/secrets.go
jonathankingfc: This is used if clair wants to use custom tls.

ricardomaraschini (approved):

PR #696, opened 2022-04-22
hpa: don't stomp replicas when managed (PROJQUAY-3647)

ricardomaraschini:
We should not set the number of replicas when the HPA is enabled. Prior
to this patch we were in a situation where the HPA was setting the
number of replicas to a given value and we were bringing it back to the
default for the deployment.


flavianmissi (approved):
LGTM

PR #698, opened 2022-04-26
quota: enable by default (PROJQUAY-3513)

ricardomaraschini:
Enable quota management feature by default.


ricardomaraschini:
@HammerMeetNail is this enough to get quota working ? I am worried as this will enable quotas for everybody (including people upgrading).

HammerMeetNail:
We decided after talking with PM that this will not be enabled by default. Can be closed.

PR #699, opened 2022-04-29
certs-and-proxies: mounting certs in mirror and renaming operator service reference (PROJQUAY-3599)

ricardomaraschini:
As per each individual commit:

**annotations: stop setting annotations on clair db deployment (PROJQUAY-3599)**

These annotations make the database to be rolled out if one of the
annotations is change. As Clair database is not affected by these
annotations we better not set them (as we already do with the Quay
database).

**cert: mount the user provided certs on mirror (PROJQUAY-3599)**

If we don't mount quay-config-tls then mirror is not aware of the
cluster wildcard cert and can't access Quay through its route.

This commits mounts the cluster wildcard cert (or the cert manually
provided by the user) in the extra_ca_certs directory.

**proxy: use the full service name (PROJQUAY-3599)**

If we don't append the "full domain" to the service name it is quite
hard for the users to allow direct traffic to quay operator service
during reconfigure.

We want to allow users to bypass traffic to svc.cluster.local.


jonathankingfc (approved):
LGTM

PR #707, opened 2022-05-09
upgrade: run upgrade job when dburi changes (PROJQUAY-3776)

ricardomaraschini:
We also need to run the upgrade job when the dburi changes, not only
when there is an operator version upgrade.


flavianmissi (dismissed):

flavianmissi:
oh darn, tests are failling

ricardomaraschini:
> oh darn, tests are failling

yeah, as now we run the upgrade job also when the database configuration changes I had to adjust the tests accordingly. PTAL

flavianmissi (approved):

PR #714, opened 2022-05-26
WIP configeditor: make config editor a managed component

nichitagutu:
This PR makes config editor a manageable component, so it can be unmanaged and deleted
([PROJQUAY-2763](https://issues.redhat.com/browse/PROJQUAY-2763))


ibazulic:
What would making the config editor a managed component and subsequent removing accomplish?

nichitagutu:
> What would making the config editor a managed component and subsequent removing accomplish?

> While password-protected, the config editor presents a security risk being exposed on the network. It's not needed for running Quay after the initial deployment nor by the operator, so it should be possible to disable the config editor or at least remove the public ingress point.
 [projquay-2763](https://issues.redhat.com/browse/PROJQUAY-2763)

My understanding is that by making config editor manageable, user will have ability to unmanage this component and delete it, so it would not pose a security risk. Correct me if I'm wrong.

kustomize/components/configeditor/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+# Clair component adds Clair v4 security scanner and its database.
```
flavianmissi: left from copy/paste?

flavianmissi (commented):
left one comment - lgtm otherwise

kustomize/components/configeditor/kustomization.yaml
nichitagutu: Indeed :/ 
Fixed

flavianmissi:
Looks like the tests need some love as well @Nikita757

pkg/kustomize/kustomize_test.go
```diff
@@ -241,9 +241,7 @@ func TestFlattenSecret(t *testing.T) {
 var quayComponents = map[string][]client.Object{
 	"quay": {
 		&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "quay-app"}},
-		&appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "quay-config-editor"}},
 		&corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "quay-app"}},
-		&corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "quay-config-editor"}},
 		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "quay-config-secret"}},
```
flavianmissi, line 245: Should this secret also be moved to the `configeditor` component?

apis/quay/v1/quayregistry_types.go
```diff
@@ -150,6 +152,7 @@ const (
 	ComponentMirrorReady        ConditionType = "ComponentMirrorReady"
 	ComponentMonitoringReady    ConditionType = "ComponentMonitoringReady"
 	ComponentTLSReady           ConditionType = "ComponentTLSReady"
+	ComponenntConfigEditorReady ConditionType = "ComponentConfigEditorReady"
```
flavianmissi, line 155: you have a typo in the const name, replace `ComponenntConfigEditorReady` with `ComponentConfigEditorReady`.
furthermore, where is this condition being used? maybe we need something in `pkg/cmpstatus` to control it?

PR #716, opened 2022-05-30
hack: add script to build and publish the operator index

flavianmissi:
also use it in the build-and-publish github workflow.

---

Test build https://github.com/flavianmissi/quay-operator/runs/6653959789?check_suite_focus=true
(it's red because the notification to slack fails).
I've deployed the operator built with the new script with `hack/deploy.sh` and ran the kuttle e2e tests against that and it worked 🎉 .

### DIFF
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -1,3 +1,5 @@
+# NOTE: set the REGISTRY secret to your own quay.io account in your fork to
+# test this workflow!
 ---
 name: Build and Publish Images
 
@@ -24,8 +26,6 @@ jobs:
     name: Set version from branch name
     env:
       BRANCH_PREFIX: redhat- # IMPORTANT! this must match the .on.push.branches prefix!
-      REGISTRY: quay.io/projectquay
-      REPO_NAME: ${{ github.event.repository.name }}
       TAG_SUFFIX: -unstable
     outputs:
       tag: ${{ steps.format-tag.outputs.tag }}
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-publish.yaml
```diff
@@ -21,17 +23,15 @@ on:
...
       BRANCH_PREFIX: redhat- # IMPORTANT! this must match the .on.push.branches prefix!
-      REGISTRY: quay.io/projectquay
-      REPO_NAME: ${{ github.event.repository.name }}
       TAG_SUFFIX: -unstable
     outputs:
       tag: ${{ steps.format-tag.outputs.tag }}
     runs-on: 'ubuntu-latest'
     steps:
-      - name: Format version
+      - name: format version
```
dmage: Why? The first step from GitHub is named `Set up job`, so for consistency it's better to keep the first letter capitalized.

.github/workflows/build-and-publish.yaml
```diff
@@ -42,191 +42,50 @@ jobs:
...
-          ref: ${{ github.event.inputs.branch || github.ref_name }}
+          wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O /usr/local/bin/yq
+          chmod +x /usr/local/bin/yq
 
-      - name: Login to Quay.io
+      - name: login to Quay.io
         uses: docker/login-action@v1
         with:
           registry: quay.io
           username: ${{ secrets.QUAY_USER }}
```
dmage, line 83: `REGISTRY` implies that it can be configured, but these variables imply that the registry have to be `quay.io`.

hack/build.sh
```diff
@@ -0,0 +1,110 @@
...
+docker build -t "${REGISTRY}/quay-operator:${TAG}" .
+docker push "${REGISTRY}/quay-operator:${TAG}"
+digest "${REGISTRY}/quay-operator:${TAG}" OPERATOR_DIGEST
+
+digest "${REGISTRY}/quay:${TAG}" QUAY_DIGEST
+digest "${REGISTRY}/clair:nightly" CLAIR_DIGEST
+digest "${REGISTRY}/quay-builder:${TAG}" BUILDER_DIGEST
+digest "${REGISTRY}/quay-builder-qemu:main" BUILDER_QEMU_DIGEST
+# shellcheck disable=SC2034
+POSTGRES_DIGEST='centos/postgresql-10-centos7@sha256:de1560cb35e5ec643e7b3a772ebaac8e3a7a2a8e8271d9e91ff023539b4dfb33'
```
dmage, line 70: wow... do we have to use this exact image?

dmage (commented):
Few questions, otherwise it looks good.

.github/workflows/build-and-publish.yaml
flavianmissi: I was aiming for consistently using lowercase, but I guess I missed `Set up job`

.github/workflows/build-and-publish.yaml
flavianmissi: true. I think we should have `REGISTRY` and `NAMESPACE`, so that both are configurable.
wdyt?

hack/build.sh
flavianmissi: it was taken from the postgres deployment: https://github.com/quay/quay-operator/blob/master/kustomize/components/postgres/postgres.deployment.yaml?plain=1#L35

.github/workflows/build-and-publish.yaml
flavianmissi: done - capitalized everything

.github/workflows/build-and-publish.yaml
flavianmissi: done - went with my own suggestion

dmage (approved):

PR #717, opened 2022-06-01
workflows: trigger build job on schedule (PROJQUAY-3871)

flavianmissi:
The schedule on the `buil-and-publish.yaml` workflow doesn't work because the workflow itself expects to run from a `redhat-*` branch, and that doesn't happen on scheduled workflows (they run from the main branch).

This PR introduces a new workflow, that triggers `buil-and-publish.yaml` via api targeting the correct branch.

I've tested this on my fork, and although the builds themselves failed (due to quay.io credentials missing), the trigger itself worked. See https://github.com/flavianmissi/quay-operator/runs/6687638116 and https://github.com/flavianmissi/quay-operator/actions/runs/2420732153 (you can inspect the output on the first link for the triggered workflow id and check that it matches the id on the second link's URL, proving that the first job triggered the second).

### DIFF
--- a/.github/workflows/build-schedule.yaml
+++ b/.github/workflows/build-schedule.yaml
@@ -1,0 +1,66 @@
+---
+name: Build and Publish Images (scheduled)
+on:
+  schedule:
+    - cron: '0 3 * * *' # run before e2e-nightly to ensure a fresh operator build
+
+jobs:
+  trigger-3.6-build:
+    name: "Build and publish 3.6"
+    runs-on: ubuntu-latest
+    env:
+      BRANCH: redhat-3.6
+    steps:
+    - uses: convictional/trigger-workflow-and-wait@v1.3.0
+      with:
+        owner: ${{ github.repository_owner }}
+        repo: 'quay-operator'
+        # see https://docs.github.com/en/actions/security-guides/automatic-token-authentication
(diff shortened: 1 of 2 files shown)


.github/workflows/build-schedule.yaml
```diff
@@ -0,0 +1,26 @@
...
+        repo: 'quay-operator'
+        # see https://docs.github.com/en/actions/security-guides/automatic-token-authentication
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        workflow_file_name: build-and-publish.yaml
+        ref: ${{ env.BRANCH }}
+        wait_interval: 30
+        inputs: |
+          {
+            "branch": "${{ env.BRANCH }}"
+          }
```
ricardomaraschini, line 26: What do you think about creating two other triggers ? I mean, one for `redhat-3.7` and another for `redhat-3.8` (assuming that we are already keeping the `redhat-3.8` in sync with `master`) ?

flavianmissi: Good idea, will do!

flavianmissi: Done, PTAL

ricardomaraschini (approved):
me approooooves

PR #722, opened 2022-06-24
clair: limit db connections pool (PROJQUAY-2659)

ricardomaraschini:
Restrict the max number of open connections to the clair postgres. If
HPA is managed we have control over the max number of pods so we can
calculate the max number of connections per pod. If HPA is not managed
then we can't properly estimate the max number of pods so we decrease
the connection pool size.

### DIFF
--- a/pkg/kustomize/secrets.go
+++ b/pkg/kustomize/secrets.go
@@ -384,12 +384,22 @@ func componentConfigFilesFor(log logr.Logger, qctx *quaycontext.QuayRegistryCont
 
 // clairConfigFor returns a Clair v4 config with the correct values.
 func clairConfigFor(log logr.Logger, quay *v1.QuayRegistry, quayHostname, preSharedKey string, configFiles map[string][]byte) ([]byte, error) {
+	// the default number for the clair's database connections pool is arbitralily defined to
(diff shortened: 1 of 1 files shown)


pkg/kustomize/secrets.go
```diff
@@ -406,9 +416,8 @@ func clairConfigFor(log logr.Logger, quay *v1.QuayRegistry, quayHostname, preSha
 			"migrations":             true,
 		},
 		"matcher": map[string]interface{}{
-			"connstring":    dbConn,
-			"max_conn_pool": 100,
```
dmage, line 410: How do we decide what should be in this map and what should be in the connstring string?

ricardomaraschini: @quay/clair should be able to shine some light here

hdonnay: Clair's config is documented (via the go types) [here](https://pkg.go.dev/github.com/quay/clair/config#Config). In this specific instance, we [deprecated](https://pkg.go.dev/github.com/quay/clair/config#Matcher) the dedicated connection pool size to make it uniformly specified in the connection string.

dmage (approved):

PR #732, opened 2022-09-07
reconciler: Do not infer clairpostgres state from config editor (PROJQUAY-4328)

jonathankingfc:
- The reconfigure endpoint is setting ClairPostgres as managed if it does not contain configuration values for it
- This should not happen here, since the config editor does not expose this configuration


harishsurf (approved):
lgtm

PR #734, opened 2022-10-03
route: Increase Quay route timeout to 60s (PROJQUAY-4002)

jonathankingfc:
- Increase Quay route timeout from default (30s) to 60s


HammerMeetNail (approved):
LGTM

PR #736, opened 2022-10-06
cleanup: Remove QuayEcosystem code (PROJQUAY-1886)

jonathankingfc:
- Remove QuayEcosystem references in the Operator
- QuayEcosystem controller is no longer available


harishsurf (approved):
LTGM, nothing legacy stuff popped out. For [upgrades.md](https://github.com/quay/quay-operator/blob/master/docs/upgrades.md#supported-quayecosystem-configurations-for-upgrades), like you said we could specify only 3.6 version supports migration

PR #741, opened 2022-11-10
olm: Mark 3.4 channel deprecated (PROJQUAY-3831)

jonathankingfc:
- Update description to mark as deprecated


HammerMeetNail:
We'll look at this later when there's a better way to deprecate old channels.

PR #746, opened 2023-02-08
WIP: Add ppc64le support to Quay Operator

jcho02:
Signed-off-by: Jason Cho <jason.cho2@ibm.com>

ppc64le support

### DIFF
--- a/bundle/manifests/quay-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/quay-operator.clusterserviceversion.yaml
@@ -156,6 +156,8 @@ spec:
                   value: quay.io/projectquay/quay-builder-qemu:main
                 - name: RELATED_IMAGE_COMPONENT_POSTGRES
                   value: centos/postgresql-13-centos7@sha256:71b24684d64da46f960682cc4216222a7e4ed8b1a31dd5a865b3e71afdea20d2
+                - name: RELATED_IMAGE_COMPONENT_POSTGRES_UPGRADE
+                  value: centos/postgresql-12-centos7@sha256:be8803d45d64870f8dfd018f3110af62e2e1558d64191faea461005e1bd03243
                 - name: RELATED_IMAGE_COMPONENT_REDIS
                   value: centos/redis-32-centos7@sha256:06dbb609484330ec6be6090109f1fa16e936afcf975d1cbc5fff3e6c7cae7542
               serviceAccountName: quay-operator
--- a/controllers/quay/quayregistry_controller_test.go
+++ b/controllers/quay/quayregistry_controller_test.go
@@ -674,7 +674,7 @@ func Test_hasNecessaryConfig(t *testing.T) {
 			quay: quayWithUnmanagedComponents(v1.ComponentClairPostgres),
 		},
 		{
-			name:   "managed clair with unmanaged clairpostgres without	 config",
+			name:   "managed clair with unmanaged clairpostgres without config",
 			experr: true,
 			cfg:    map[string][]byte{},
 			quay:   quayWithUnmanagedComponents(v1.ComponentClairPostgres),
--- a/docs/image-overrides.md
+++ b/docs/image-overrides.md
@@ -1,6 +1,6 @@
 # Image Overrides
 
-The container images used by the managed components deployed by the Quay Operator can be overridden. 
+The container images used by the managed components deployed by the Quay Operator can be overridden.
 
 **NOTE**: This should only be done for development, testing, and debugging as it not guaranteed that all components will be compatible when overriding the defaults included in the Operator.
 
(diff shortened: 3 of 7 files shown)


.github/workflows/build-and-publish.yaml
```diff
@@ -70,7 +70,11 @@ jobs:
       - name: Install yq
         env:
           VERSION: v4.14.2
-          BINARY: yq_linux_amd64
+          override-build: |
```
dmage: `override-build` is not mentioned by [GitHub Actions Docs](https://docs.github.com/en/search?query=%22override-build%22), what is it?

jcho02: Updated the workflow files

.github/workflows/build-and-publish.yaml
```diff
@@ -67,14 +67,24 @@ jobs:
...
+        if: ${{ "$(uname -m)" == "amd64" }}
         env:
           VERSION: v4.14.2
           BINARY: yq_linux_amd64
         run: |
           wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O /usr/local/bin/yq
           chmod +x /usr/local/bin/yq
 
+      - name: Install yq for ppc64le
+        if: ${{ "$(uname -m)" == "ppc64le" }}
```
dmage: We use GitHub-hosted runners, and they are available [only for x86_64](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

Also you cannot use shell commands in GitHub expressions `${{ ... }}`, see [GitHub Docs for Expressions](https://docs.github.com/en/actions/learn-github-actions/expressions).

Dockerfile
```diff
@@ -10,7 +10,11 @@ COPY apis/ apis/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod vendor -o manager main.go
+RUN if [ ${ARCH}="ppc64le" ] ; then \
```
dmage: You can simply remove `GOOS=linux GOARCH=amd64`.

Makefile
```diff
@@ -19,7 +26,7 @@ test: generate fmt vet manifests
 
 test-e2e:
 	mkdir -p ./bin
-	curl -L https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_x86_64 -o ./bin/kuttl;
+	curl -L https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_${ARCH} -o ./bin/kuttl;
```
dmage: Don't complicate things, use `_linux_$(uname -m)`

dmage:
Did you try to build the docker image on ppc64le?

.github/workflows/build-and-publish.yaml
jcho02: How would you like us to proceed regarding workflows if GH runners are only available for x86_64.

Dockerfile
modassarrana89: @dmage why can't we use --platform concept . 
```
FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.17 as builder
ARG BUILDPLATFORM 
ARG TARGETARCH
RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on go build -mod vendor -o manager main.go
```

I am actually planning to support quay-operator on Z & doing the same changes 

For makefile
------------
```
export DOCKER_CLI_EXPERIMENTAL=enabled

.PHONY: build # Build the container image
docker-build:
        @docker buildx create --use --name=crossplat --node=crossplat && \
        docker buildx build --platform linux/s390x,linux/amd64,linux/ppc64le  --tag $(IMG) .
```

Makefile
modassarrana89:
```
arch=$(shell uname -m)
...

curl -L https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_$(arch) -o ./bin/kuttl;
...
```

jcho02:
Docker image builds on ppc64le and deploys on ppc64le supported OCP clusters.
WIP tag for QuayRegistry's

Makefile
```diff
@@ -63,7 +62,7 @@ generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 # Build the docker image
-docker-build: test
+docker-build:
```
yussufsh: Why are we removing test target?

jcho02: Oops, sorry removed that while testing locally. Test target is back

PR #753, opened 2023-04-05
env: Add env variable for pg 12 (PROJQUAY-4608)

jonathankingfc:
- Add env variable for postgres 12

### DIFF
--- a/pkg/kustomize/kustomize.go
+++ b/pkg/kustomize/kustomize.go
@@ -91,6 +91,33 @@ func ComponentImageFor(component v1.ComponentKind) (types.Image, error) {
 	return imageOverride, nil
 }
 
+// ComponentImageFor checks for an environment variable indicating which component container image
+// to use. If set, returns a Kustomize image override for the given component.
+func postgresUpgradeImage() (types.Image, error) {
+	imageOverride := types.Image{
+		Name: "centos/postgresql-12-centos7",
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/kustomize/kustomize.go
```diff
@@ -91,6 +91,33 @@ func ComponentImageFor(component v1.ComponentKind) (types.Image, error) {
 	return imageOverride, nil
 }
 
+// ComponentImageFor checks for an environment variable indicating which component container image
+// to use. If set, returns a Kustomize image override for the given component.
+func postgresUpgradeImage() (types.Image, error) {
+	imageOverride := types.Image{
+		Name: "centos/postgresql-13-centos7",
```
dmage: I believe it should be `centos/postgresql-12-centos7`

pkg/kustomize/kustomize.go
```diff
@@ -471,6 +498,12 @@ func KustomizationFor(
 		}
 	}
 
+	if pgImage, err := postgresUpgradeImage(); err != nil {
+		return nil, err
+	} else {
```
dmage, line 503: You don't need `} else {` after `return`

pkg/kustomize/kustomize.go
jonathankingfc: Fixed

dmage (approved):

PR #755, opened 2023-05-09
fix: Use autoscaling/v2 (PROJQUAY-5079)

dmage:
autoscaling/v2beta2 isn't available on OCP 4.13, so we have to migrate to autoscaling/v2.


jonathankingfc (approved):
LGTM

PR #757, opened 2023-05-12
ppc64le support for Quay Operator (PROJQUAY-5372)

jcho02:
Builds and deploys successfully on ppc64le OCP clusters by running `./hack/build.sh` and then `./hack/deploy.sh`
e2e tests pass for all quay-registry builds

Dockerfile supports amd64, ppc64le image operator builds
Makefile runs e2e tests for amd64, ppc64le environments
build.sh has updated amd64, ppc64le supported tag dependencies (quay, quay-builder, quay-builder-qemu clair, redis, postgres)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,9 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     container: docker.io/library/golang:latest
+    strategy:
+      matrix:
+        platform: ['amd64','ppc64le']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


hack/build.sh
```diff
@@ -65,13 +65,13 @@ digest "${REGISTRY}/${NAMESPACE}/quay-operator:${TAG}" OPERATOR_DIGEST
 digest "${REGISTRY}/${NAMESPACE}/quay:${TAG}" QUAY_DIGEST
 digest "${REGISTRY}/${NAMESPACE}/clair:nightly" CLAIR_DIGEST
 digest "${REGISTRY}/${NAMESPACE}/quay-builder:${TAG}" BUILDER_DIGEST
-digest "${REGISTRY}/${NAMESPACE}/quay-builder-qemu:main" BUILDER_QEMU_DIGEST
+digest "${REGISTRY}/${NAMESPACE}/quay-builder-qemu:3.9.0" BUILDER_QEMU_DIGEST
 # shellcheck disable=SC2034
-POSTGRES_DIGEST='centos/postgresql-13-centos7@sha256:71b24684d64da46f960682cc4216222a7e4ed8b1a31dd5a865b3e71afdea20d2'
+POSTGRES_DIGEST='quay.io/sclorg/postgresql-15-c9s@sha256:593910f2d4b895f4924261a3b8b2aa6457892100a01a0c0ad661cd378d810d65'
```
dmage: Quay 3.9 users will use Postgres 13, we cannot switch to Postgres 15 unless we provide automated migration of the data format from 13 to 15.

jcho02: Changed `POSTRGES_DIGEST` to postgres 13. Left `POSTGRES_UPGRADE_DIGEST`to what it was previously. (doesn't seem to be used by p).

jcho02:
Updated dockerfile, build.sh and image.yaml to support multi-arch builds.
ci.yaml edited build phase for ppc64le builds, removed test for ppc64le as it needs a cluster, tests run successfully locally with correct dependencies.
Removed `s390x` tag to let Z test and add themselves

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:1.19 as builder
```
dmage, line 1: BUILDPLATFROM is the host system platform, i.e. usually it's linux/amd64.

The current version of Dockerfile will create binaries for the host system, not the target system. Either you need to remove `--platform=$BUILDPLATFORM` and rely on emulation, or use TARGETARCH and TARGETOS and rely Go ability to cross-compile.

```
FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi8/go-toolset:1.19 AS build
...
ARG TARGETOS TARGETARCH
RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -mod vendor -o manager main.go
```

See https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/ for details.

hack/build.sh
```diff
@@ -58,20 +58,19 @@ function digest() {
...
 digest "${REGISTRY}/${NAMESPACE}/quay-operator:${TAG}" OPERATOR_DIGEST
 
 digest "${REGISTRY}/${NAMESPACE}/quay:${TAG}" QUAY_DIGEST
 digest "${REGISTRY}/${NAMESPACE}/clair:nightly" CLAIR_DIGEST
 digest "${REGISTRY}/${NAMESPACE}/quay-builder:${TAG}" BUILDER_DIGEST
-digest "${REGISTRY}/${NAMESPACE}/quay-builder-qemu:main" BUILDER_QEMU_DIGEST
+digest "${REGISTRY}/${NAMESPACE}/quay-builder-qemu:3.9.0" BUILDER_QEMU_DIGEST
 # shellcheck disable=SC2034
-POSTGRES_DIGEST='centos/postgresql-13-centos7@sha256:71b24684d64da46f960682cc4216222a7e4ed8b1a31dd5a865b3e71afdea20d2'
+POSTGRES_DIGEST='quay.io/sclorg/postgresql-13-c9s@sha256:593910f2d4b895f4924261a3b8b2aa6457892100a01a0c0ad661cd378d810d65'
```
dmage, line 69: @jonathankingfc will upgrades work with this image? won't it break downstream?

Dockerfile
jcho02: Updated dockerfile to have `TARGETOS` and `TARGETARCH` as args

hack/build.sh
jonathankingfc: the downstream images are defined in the midstream cpaas repo, so this won't affect the downstream builds

jonathankingfc (dismissed):
LGTM

jcho02:
Finalized build.sh to have proper multi-arch build for operator-index image. 
A single-arch operator-index image will be built bundled by the correct operator-bundle image before manifesting into one multi-arch operator-index image.

.github/workflows/ci.yaml
```diff
@@ -30,16 +33,20 @@ jobs:
...
+      matrix:
+        go-version: ['1.20']
+        platform: ['amd64','ppc64le']
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: OS Dependencies
-        run: apt-get update && apt-get install -y tar make gcc
+        run: apt-get update && apt-get install -y tar make gcc docker.io
```
yussufsh, line 49: Is `docker.io` needed now?

Dockerfile
yussufsh: @jcho02 need to remove `--platform=$BUILDPLATFORM`?

.github/workflows/ci.yaml
dmage: we can remove it in a follow-up PR

Dockerfile
jcho02: no

jcho02: no

PR #764, opened 2023-06-01
pgupgrade: Scale back clair pods when upgrading clair db (PROJQUAY-5235)

jonathankingfc:
- When upgrading the clair postgres, scale back the clair pods to prevent further operations
- This will cause the previous postgres deployment to terminate and the upgrade can complete


harishsurf (approved):
lgtm

PR #768, opened 2023-06-12
chore: fix multiarch tests (PROJQUAY-5382)

yussufsh:
Existing tests were not running on correct arch, used setup-qemu-action. Removed Z from test matrix as the kubebuilder and other dependencies are not available. Upgraded all kubebuilder, etcd, ctl, apiserver to the latest version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,21 +37,30 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


yussufsh:
@dmage ptal

yussufsh:
We will not run CI for ppc64le or s390x because of the dependencies support. As agreed we do not need to run unit tests, only amd64 will suffice.
Will create a different PR for removing unwanted jobs from CI.yaml which are not running on the assumed arch.

PR #775, opened 2023-06-20
pgupgrade: Change restart policy to Never to preserve error message (PROJQUAY-5235) (5361)

jonathankingfc:
- Change restart policy to Never to preserve error message


harishsurf (approved):
Take a look at [this limitation](https://docs.openshift.com/container-platform/4.9/nodes/jobs/nodes-nodes-jobs.html#jobs-limits_nodes-nodes-jobs). It looks like `restartsPolicy:never` still does a pod restart but via `Job` controller. Not sure if we wold still be able to see the actual pod failure for `BackoffLimitExceeded` error

PR #776, opened 2023-06-21
pgupgrade: Decrease termination grace period to pod shuts down sooner (PROJQUAY-5631)

jonathankingfc:
- Termination grace period of 180 seconds is too long, causes delays in shutdown

### DIFF
--- a/kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
+++ b/kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
@@ -9,7 +9,7 @@ metadata:
 spec:
   template:
     spec:
-      restartPolicy: Never
+      restartPolicy: OnFailure
       terminationGracePeriodSeconds: 180
       serviceAccountName: clair-postgres
       volumes:
@@ -95,4 +95,4 @@ spec:
             - "run-postgresql"
           args:
             - "--version"
-  backoffLimit: 20
+  backoffLimit: 50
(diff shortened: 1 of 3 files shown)


kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 spec:
   template:
     spec:
-      restartPolicy: Never
+      restartPolicy: OnFailure
```
dmage, line 12: Are you sure you won't need logs from previous attempts?

jonathankingfc: This will only work for debugging, for production we need to restart the job if failed

dmage: `restartPolicy: Never` means that the pod should be re-created by the job controller, not restarted by kubelet.
With `restartPolicy: Never` we potentially will have 50 (failed) pods with their logs. With `restartPolicy: OnFailure` we will only have the logs of the last attempt and the previous one. From QE results, the first attempt has important logs that we shouldn't loose.

kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
```diff
@@ -95,4 +95,4 @@ spec:
             - "run-postgresql"
           args:
             - "--version"
-  backoffLimit: 20
+  backoffLimit: 50
```
dmage, line 98: If the problem is that kubelet is killing postgres without giving it a chance to delete its pid file, increasing backoff limit won't help — the job doesn't try to recover from an ungraceful postgres shutdown.

jonathankingfc:
No longer needed

PR #777, opened 2023-06-26
update index images to support multi-arch builds  (PROJQUAY-5693)

jcho02:
Index image builds weren't building based off their arch's. All 3 images were being built as x86.
`--binary-image` tag with opm image based off their arch.
Tested and deployed on OCP cluster.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -115,11 +115,11 @@ POWER_DIGEST=$(skopeo inspect --raw  docker://${REGISTRY}/${NAMESPACE}/quay-oper
 Z_DIGEST=$(skopeo inspect --raw  docker://${REGISTRY}/${NAMESPACE}/quay-operator-bundle:${TAG} | \
            jq -r '.manifests[] | select(.platform.architecture == "s390x" and .platform.os == "linux").digest')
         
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -115,11 +115,11 @@ POWER_DIGEST=$(skopeo inspect --raw  docker://${REGISTRY}/${NAMESPACE}/quay-oper
 Z_DIGEST=$(skopeo inspect --raw  docker://${REGISTRY}/${NAMESPACE}/quay-operator-bundle:${TAG} | \
            jq -r '.manifests[] | select(.platform.architecture == "s390x" and .platform.os == "linux").digest')
         
-opm index add --build-tool docker --bundles "${REGISTRY}/${NAMESPACE}/quay-operator-bundle@${AMD64_DIGEST}" --tag "${REGISTRY}/${NAMESPACE}/quay-operator-index:${TAG}-amd64"
+opm index add --build-tool docker --bundles "${REGISTRY}/${NAMESPACE}/quay-operator-bundle@${AMD64_DIGEST}" --tag "${REGISTRY}/${NAMESPACE}/quay-operator-index:${TAG}-amd64" --binary-image "quay.io/operator-framework/opm:latest@sha256:0b10c3c3f713f11c805505e0aa686ce2a347b814814c5e110a084903e513babe"
```
yussufsh: Can we use a fixed version here instead of latest? with tag names for archs.

PR #778, opened 2023-06-27
pgupgrade: Use remote migration from SCL images (PROJQUAY-5631)

jonathankingfc:
- Switch to rhel8 images for more efficient implementation


controllers/quay/quayregistry_controller.go
```diff
@@ -242,6 +243,38 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
...
+					Namespace: quay.GetNamespace(),
+				},
+				oldPostgresDeployment,
+			); err != nil {
+				r.Log.Info(fmt.Sprintf("%s deployment not found, skipping", oldPostgresDeploymentName))
+				continue
+			}
+
+			// Remove owner reference
+			obj, err := v1.RemoveOwnerReference(quay, oldPostgresDeployment)
```
dmage, line 266: Why do you need to remove owner references before calling Delete?

pkg/middleware/middleware.go
```diff
@@ -115,7 +115,7 @@ func Process(quay *v1.QuayRegistry, obj client.Object, skipres bool) (client.Obj
 				// if no number of replicas has been set in kustomization files
 				// we set its value to two or to the value provided by the user
 				// as an override (if provided).
-				desired := pointer.Int32(2)
```
jonathankingfc: Remove and include in future PR

kustomize/components/clairpostgres/postgres.deployment.yaml
```diff
@@ -26,9 +26,9 @@ spec:
             name: clair-postgres-conf-sample
         - name: postgres-data
           persistentVolumeClaim:
-            claimName: clair-postgres
+            claimName: clair-postgres-13
       containers:
-        - name: postgres
+        - name: postgres-upgrade
```
jonathankingfc: Change this name, this is incorrect. This should just be postgres-clair

controllers/quay/quayregistry_controller.go
```diff
@@ -242,6 +243,38 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
...
+			if err != nil {
+				log.Error(err, "could not remove owner reference from old postgres deployment")
+			}
+
```
jonathankingfc, line 270: // remove owner reference to clair-pg-old.pvc

controllers/quay/quayregistry_controller.go
```diff
@@ -242,6 +243,61 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
 
 		if job.Status.Succeeded == 1 {
 			log.Info(fmt.Sprintf("%s upgrade complete", jobName))
+			var oldPostgresDeploymentName string
+			if jobName == clairPostgresUpgradeJobName {
+				oldPostgresDeploymentName = fmt.Sprintf("%s-%s", quay.GetName(), "clair-postgres-old")
+			} else {
+				oldPostgresDeploymentName = fmt.Sprintf("%s-%s", quay.GetName(), "quay-postgres-old")
```
harishsurf: nit: should these `-old` string literal be defined as `const`?

kustomize/components/pgupgrade/quay-pg-old.deployment.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: quay-database-old
```
harishsurf, line 4: shouldn't this be `quay-postgres-old` as defined in [checkPostgresUpgradeStatus()](https://github.com/quay/quay-operator/pull/778/files#diff-db06dd075ea792819f15dcbfb9c2376eea2e17832c2bd64ae6b381d3c947b57eR250)

controllers/quay/quayregistry_controller.go
```diff
@@ -242,6 +243,61 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
...
+				ctx,
+				obj,
+			); err != nil {
+				r.Log.Error(err, fmt.Sprintf("%s deployment could not be deleted", oldPostgresDeploymentName))
+			}
+
+			// Remove owner reference from old pvc so user can delete when ready
+			var oldPostgresPVCName string
+			if jobName == clairPostgresUpgradeJobName {
+				oldPostgresPVCName = fmt.Sprintf("%s-%s", quay.GetName(), "clair-postgres")
```
harishsurf, line 282: nit: could we use similar PVC suffix name (either one) for both clair and quay?

kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
```diff
@@ -9,84 +9,41 @@ metadata:
...
-          resources:
-            requests:
-              cpu: 500m
-              memory: 2Gi
-          command:
-            - "/bin/bash"
-            - "-c"
-            - "cp -r /var/lib/pgsql/data /var/lib/pgsql/backup/clair && run-postgresql --help"
-
+            claimName: clair-postgres-13
```
harishsurf, line 21: nit: should we name PVCs `old/new` instead of `-13` since the upgrade can be done for future pg versions too?

harishsurf (commented):
added some minor comments

controllers/quay/quayregistry_controller.go
dmage: `clair-postgres-old` is used only on this line and in yaml; if we introduce `const`, it'll make code more complex and it'll be harder to find code that interacts with `clair-postgres-old`. What benefits will it give us?

dmage (commented):
So the upgrade looks like this:

- operator scales down the existing postgres
- it creates quay-postgres-old with Postgres 10 and existing data
- it creates the upgrade job that mounts the new PVC (quay-postgres-13) and populates it from quay-postgres-old
- once the job completes the migration, the operator deletes quay-postgres-old and scales up quay-database

Is my understanding correct?

kustomize/components/clairpgupgrade/clair-pg-upgrade.job.yaml
dmage: if we are going to create a new PVC for every new Postgres version, we cannot use old/new. Today postgres-13 is "new", next release it'll be "old", and we cannot rename PVCs.

controllers/quay/quayregistry_controller.go
harishsurf: just that we don't hardcode values, but since it's only this line, it's okay to skip

controllers/quay/quayregistry_controller.go
jonathankingfc: Otherwise the operator will redeploy this object

jonathankingfc:
> 

This is correct

controllers/quay/quayregistry_controller.go
dmage: RemoveOwnerReference removes `quay` from `metadata.ownerReferences` of `oldPostgresDeployment`. Then `oldPostgresDeployment` gets deleted.

What tells the operator not to redeploy the object?

jonathankingfc: Because we do not re-rollout the pguprade component

dmage: I don't get it. How are ownerReferences related to rollouts?

It'll help if you point to the source code that uses ownerReferences during rollouts.

jonathankingfc: We are removing the owner reference so that the resource is no longer "managed" by the operator.

dmage (approved):

PR #803, opened 2023-07-25
image: Update centos7 image to pull from docker.io (PROJQUAY-5853)

jonathankingfc:
- Add docker.io registry explicit name to image


dmage (changes requested):
This value is used mainly by `hack/prepare-upstream.sh`, this script will convert it back to something like `centos/postgresql-10-centos7@sha256:...`

jonathankingfc:
No longer needed

PR #804, opened 2023-07-27
change storage.sh ODF version to user's OC version

jcho02:
`./hack/storage.sh` fails as stable-4.9 is no longer supported in OperatorHub. Using latest version 4.13 to successfully build ODF via storage.sh rather than installing manually via OperatorHub

### DIFF
--- a/hack/storage.sh
+++ b/hack/storage.sh
@@ -6,6 +6,9 @@
 #  * a valid login session to an OCP cluster, with cluster admin privileges
 #  * `oc`
 
+export TAG=${TAG:-"4"}
+VERSION=$(oc version | grep 'Client Version' | cut -f2 -d ":" | cut -f2 -d ".")
+
 # prints pre-formatted info output.
 function info {
 	echo "INFO $(date '+%Y-%m-%dT%H:%M:%S') $*"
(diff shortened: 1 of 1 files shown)


dmage:
The version of ODF must match the version of OCP that you use. On OCP 4.13 there should be only two channels be available: stable-4.13 and stable-4.12 (for upgrades). On OCP 4.12 stable-4.13 won't be available.

jcho02:
Instead of using a preset version, latest changes have the script read for cluster version and build based off cluster version. So that users won't have to manually change.

hack/storage.sh
```diff
@@ -6,6 +6,9 @@
 #  * a valid login session to an OCP cluster, with cluster admin privileges
 #  * `oc`
 
+export TAG=${TAG:-"4"}
+VERSION=$(oc version | grep 'Client Version' | cut -f2 -d ":" | cut -f2 -d ".")
```
dmage: It should be the server's version, i.e. something like

```
oc get clusterversion version -o jsonpath='{range .status.history[?(@.state=="Completed")]}{.version}{"\n"}{end}' | head -n1 | cut -d. -f1-2
```

dmage (approved):

PR #805, opened 2023-07-31
pgupgrade: Add POSTGRES_UPGRADE_CLEANUP env variable (PROJQUAY-5852)

jonathankingfc:
- Adds POSTGRES_UPGRADE_CLEANUP env variable to determine whether or not old PVCs should be deleted after a successful upgrade


harishsurf (approved):
lgtm

PR #814, opened 2023-08-12
workflow: Nightly job for running e2e on ppc64le cluster

yussufsh:


yussufsh:
@dmage please help with the review. Once approved I will add the secrets needed for running this nightly.
Going forward we may reduce the frequency.

dmage (approved):

PR #816, opened 2023-08-17
chore: remove unused arch matrix

yussufsh:
Existing jobs had multiple matrix configured but all were running on amd64. 
Following up from https://github.com/quay/quay-operator/pull/768#issuecomment-1681590048 we can remove the matrix and run tests only on amd64.


yussufsh:
@dmage ptal

dmage (approved):

PR #826, opened 2023-08-30
reconfigure: Remove ca-bundle.crt and service-ca.crt (PROJQUAY-5233)

jonathankingfc:
- We need to strip out the cluster provided certs from the reconfigure rrequest to avoid duplication.
- This resolves JSON encoding/decoding issues found in passing this cert through the config editor

### DIFF
--- a/pkg/configure/configure.go
+++ b/pkg/configure/configure.go
@@ -178,6 +178,10 @@ func createUpdatedSecret(reconfigureRequest request, oldSecret corev1.Secret) co
 
 	for fullFilePathname, encodedCert := range reconfigureRequest.Certs {
 		certName := strings.Split(fullFilePathname, "/")[len(strings.Split(fullFilePathname, "/"))-1]
+		// We must strip out cluster provisioned CA certs, as they will be mounted into the pod from separate config maps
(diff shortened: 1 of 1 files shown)


pkg/configure/configure.go
```diff
@@ -178,6 +178,10 @@ func createUpdatedSecret(reconfigureRequest request, oldSecret corev1.Secret) co
 
 	for fullFilePathname, encodedCert := range reconfigureRequest.Certs {
 		certName := strings.Split(fullFilePathname, "/")[len(strings.Split(fullFilePathname, "/"))-1]
+		// We must strip out cluster provisioned CA certs, as they will be mounted into the pod from separate config maps
+		if certName == "service-ca.crt" || certName == "ca-bundle.crt" {
```
dmage, line 182: Won't config-tool be a better place for this filtering? Why does the operator need to know about this strangeness of config-tool?

jonathankingfc: If reconfigure requests from the config editor, then this problem only exists in an operator setting. However, since we are going to most likely deprecate the config-tool, we can move it there for the time being.

jonathankingfc:
Closing in favor of https://github.com/quay/quay/pull/2231

PR #830, opened 2023-09-04
chore: fix scheduled build and publish due to breaking upgrade

yussufsh:
In the action `convictional/trigger-workflow-and-wait` the inputs paramater `inputs` was renamed to `client_payload`. Since last bump to v1.6.5 this was not changed.


yussufsh:
@dmage ptal

PR #835, opened 2023-09-06
[redhat-3.8] configmaps: Track resource versions of cluster provided CA certs (PROJQUAY-5174)

dmage:
Backport of #833.

- The config.openshift.io/inject-trusted-cabundle: 'true' annotation will inject content into the config map after kustomize has generated its resources. This means that kustomize will not be aware of changes made by Openshift to these config maps
- In order to redeploy QuayRegistry resources when cluster CA are rotated, this PR implements a mechanism to track the resource version of the CA as an annotation on the Quay resources


jonathankingfc (approved):
LGTM

PR #840, opened 2023-09-21
chore: use ubi9

dmage:


dmage:
/assign @jonathankingfc 

We are migrating to RHEL9, the operator seems to works fine with the new base image. Can you approve this PR?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi8/go-toolset:1.19 as builder
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.19 as builder
```
arborite-rh, line 1: should we change the tag to use 1.20 to match the go.mod go version? 
```suggestion
FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
```

PR #859, opened 2023-11-10
route: updating router timeout to 10m (PROJQUAY-2679)

bcaton85:
Increases timeout specifically for the case of large image pushes.


jonathankingfc (approved):
LGTM

PR #874, opened 2023-12-14
reconfigure: Use config editor secret password as credential for reconfigure request (PROJQUAY-6021)

dmage:


jonathankingfc (approved):
LGTM

PR #890, opened 2024-02-05
pgupgrade: Do not try to scale down clair when unmanaged (PROJQUAY-6457)

jonathankingfc:
- When Clair is unmanaged, do not attempt to scale it down during Clair PG upgrade


arborite-rh (approved):
lgtm

PR #891, opened 2024-02-06
clair: Use clair config merge (PROJQUAY-6577)

jonathankingfc:
- Use clair config merge method to join user provided config with Quay gnerated


pkg/kustomize/secrets.go
```diff
@@ -370,11 +370,12 @@ func componentConfigFilesFor(log logr.Logger, qctx *quaycontext.QuayRegistryCont
 			preSharedKey = config.(map[string]interface{})["SECURITY_SCANNER_V4_PSK"].(string)
 		}
 
-		cfg, err := clairConfigFor(log, quay, quayHostname, preSharedKey, configFiles)
+		cfg, err := clairConfigFor(log, quay, quayHostname, preSharedKey)
 		if err != nil {
 			return nil, err
 		}
 		cfgFiles["config.yaml"] = cfg
+		cfgFiles["config.yaml.d/user-config.yaml"] = configFiles["clair-config.yaml"]
```
hdonnay: I'd suggest using a name like `50-user-config.yaml` -- Clair merges things lexically, so using a numeric prefix makes it easy to control the order.

arborite-rh (approved):

PR #895, opened 2024-02-22
clair: Ignore unpatched vulns by default (PROJQUAY-5954)

jonathankingfc:
- In the default Clair config, set ignore_unpatched: true


crozzy (approved):
LGTM

crozzy:
Just for visibility; this isn't a simple toggle. If `ignore_unpatched` was false or omitted (and the version of Clair supported unpatched vulns) those vulnerabilities will be in the DB and will be matched against. Adding the option after the fact will not make those vulns unmatchable.

PR #899, opened 2024-03-04
clair: Fix subpath mount points in deployment (PROJQUAY-6726)

jonathankingfc:
- Add subpath to config.yaml to prevent overwriting subdirectory


harishsurf (approved):
lgtm! were you able to test this out?

jonathankingfc:
> lgtm! were you able to test this out?

Yes, this fixes the case when the user does not provide a custom config, ie that field is empty.

PR #919, opened 2024-05-29
migration: don't delete old PVC by default during upgrade (PROJQUAY-7182)

Sunandadadi:
Removing check to delete old PVC, don't delete it by default.

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -301,8 +301,9 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
 				log.Error(err, "could not remove owner reference from old postgres pvc")
 			}
 
-			// If user has set POSTGRES_UPGRADE_RETAIN_BACKUP=true, then we should not delete the old PVC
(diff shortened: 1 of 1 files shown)


controllers/quay/quayregistry_controller.go
```diff
@@ -302,9 +301,11 @@ func (r *QuayRegistryReconciler) checkPostgresUpgradeStatus(
 			}
 
 			// If user has set POSTGRES_UPGRADE_RETAIN_BACKUP=true, then we should not delete the old PVC
-			if os.Getenv("POSTGRES_UPGRADE_RETAIN_BACKUP") == "true" {
-				continue
-			}
+			//if os.Getenv("POSTGRES_UPGRADE_RETAIN_BACKUP") == "true" {
```
jonathankingfc: Instead of removing this completely, can we just reverse the flag? To something like POSTGRES_UPGRADE_DELETE_BACKUP ?

Sunandadadi: sure, that's a good idea

jonathankingfc (approved):
LGTM

PR #942, opened 2024-08-22
quay-app(Route-timeout): increasing timeout for external S3 Storage system with larger images (PROJQUAY-7670)

michaelalang:
With various new Storage Class integrations and possible delays in moving the multi-part objects to it's final destination, we've seen various timeouts being raised even though multi-parts have been uploaded successful. 
These timeout's are not happening during uploads but when the client(s) wait for Quay to complete the multi-parts which might take more than the currently defined `10m` 

It's further to notice that sharding should be considered to avoid affecting other Ingress connections which do not have such a long duration.


michaelalang:
@ibazulic wait ... there's some commit's in this PR that shouldn't be in ... trying to fix

jbpratt:
Closing, it seems this has been bumped to 30m. Thanks as always Michaela!

PR #959, opened 2024-09-20
pgupgrade: updating postgres version to 15 fort clair and quay (PROJQUAY-7253)

deshpandevlab:
pgupgrade: Upgraded postgresql version to 15 : https://github.com/sclorg/postgresql-container/tree/master/15

- This PR updates postgresql version for clair and quay

### Testing
- `make test`
- `make test-e2e`

PG version before the update

```
PostgreSQL 13.11
```

PG version after this update

```
PostgreSQL 15.8
```


ibazulic:
Change of PostgreSQL version will require a complete migration again, as we've done with PostgreSQL 10 -> PostgreSQL 13. This is because of different binary formats that the two databases have on disk, PostgreSQL 15 cannot be started on PostgreSQL 13 data directory.

bcaton85:
@jonathankingfc can you comment on this? @deshpandevlab Have you seen errors when testing upgrades from an existing quay instance?

jonathankingfc:
@ibazulic That is correct. The mechanism for upgrading from 13 to 15 is the same one as we used to upgrade from 10 to 13. Updating the env vars and the defaults as in this PR is enough.

jonathankingfc (approved):
LGTM

PR #977, opened 2024-10-11
revert: return on error if objectbucketclaim not found (PROJQUAY-8072)

bcaton85:
This prevents the install from going through.


deshpandevlab:
LGTM

PR #927, opened 2024-05-30
statefulset: Use stateful sets for posgres deployments (PROJQUAY-6672)

jonathankingfc:
- Swap Postgres and Clair Postgres Deployments to StatefulSets

## Summary by Sourcery

Convert Postgres deployments for Quay and Clair from Deployments to StatefulSets to improve database management and persistence

Enhancements:
- Migrate Postgres and Clair Postgres configurations from Deployments to StatefulSets to provide better stateful application management

CI:
- Update CI workflow to simplify test command
- Remove manual Kubebuilder installation step

Chores:
- Update Kustomize configurations and patches to reflect StatefulSet changes
- Modify Kubernetes resource management to use StatefulSet kind

PR #980, opened 2024-10-16
statefulset: Use stateful sets for postgres deployments (PROJQUAY-6672)

deshpandevlab:
We're adding a check for Deployment here because during the migration process from Deployment to StatefulSet for the Clair Postgres component, we cannot assume that the StatefulSet will always be present. There may be instances where the old Deployment is still in use, especially if the upgrade process is ongoing or hasn't been initiated yet. This ensures that we can handle both the old Deployment-based setup and the new StatefulSet-based setup, providing a smoother transition and backwards compatibility during the upgrade process. 

### Testing
- e2e
- manually validated the migration

```
oc get pods -n shubhra
NAME                                              READY   STATUS             RESTARTS       AGE
shubhra-registry-clair-app-84b66cf595-tfm2p       1/1     Running            0              2m12s
shubhra-registry-clair-app-84b66cf595-zr64h       1/1     Running            0              2m12s
shubhra-registry-clair-postgres-0                 1/1     Running            0              2m12s
shubhra-registry-clair-postgres-bbf8d4558-hwd7q   1/1     Terminating        0              10m
shubhra-registry-clair-postgres-upgrade-plmnn     0/1     Completed          0              5m36s
shubhra-registry-quay-app-596b9db6c-6nmsh         1/1     Running            1 (13m ago)    13m
shubhra-registry-quay-app-596b9db6c-j9nzh         1/1     Running            2 (13m ago)    13m
shubhra-registry-quay-app-858546c478-5sxcs        0/1     CrashLoopBackOff   5 (2m4s ago)   5m36s
shubhra-registry-quay-app-upgrade-px67m           0/1     Completed          0              14m
shubhra-registry-quay-database-5b6d4b9897-t4x2w   1/1     Running            0              13m
shubhra-registry-quay-redis-76bc4fd7f4-x8l4s      1/1     Running            0              13m
```


ibazulic:
> We're adding a check for Deployment here because during the migration process from Deployment to StatefulSet for the Clair Postgres component, we cannot assume that the StatefulSet will always be present. There may be instances where the old Deployment is still in use, especially if the upgrade process is ongoing or hasn't been initiated yet. This ensures that we can handle both the old Deployment-based setup and the new StatefulSet-based setup, providing a smoother transition and backwards compatibility during the upgrade process.

There should never be a case where we need to manage both deployments and statefulsets at the same time IMHO. Once the operator sees a deployment, it should shut down all Quay instances immediately and begin a migration from a deployment to a statefulset. There should be no instances of Quay or Clair running during the migration period, no jobs, no activity, apart from the migration pod.

deshpandevlab:
> There should never be a case where we need to manage both deployments and statefulsets at the same time IMHO. Once the operator sees a deployment, it should shut down all Quay instances immediately and begin a migration from a deployment to a statefulset. There should be no instances of Quay or Clair running during the migration period, no jobs, no activity, apart from the migration pod.


The check for both StatefulSet and Deployment is not intended to be about managing both simultaneously or during database migration. It's specifically to handle the transition period when upgrading the operator itself from a version that used Deployments to one that uses StatefulSets for Clair Postgres.

This allows the operator to correctly identify the current state (Deployment or StatefulSet) and proceed with the appropriate upgrade path, ensuring a smooth transition between operator versions without assuming a specific resource type exists.

The actual migration process from Deployment to StatefulSet would still occur as a separate, controlled step as you described.

ibazulic:
@deshpandevlab ah, gotcha, so it's only used for identification of the current state. Thanks! I misunderstood what the actual purpose is.

crozzy (approved):
Conceptually it LGTM, I can't offer an in-depth code review as I'm not so familiar with the operator.

I assume this PR supersedes https://github.com/quay/quay-operator/pull/927/files?

deshpandevlab:
/test ocp-latest-e2e

test failed for mounting error. retrying the test

PR #982, opened 2024-10-18
pgupgrade: updating pvc reference (PROJQUAY-8092)

bcaton85:
The "old" clair postgres deployment needs to refer to the correct PVC. In this case the upgrade was referring to `clair-postgres` instead of `clair-postgres-13` which caused the migration to run with no data.


jonathankingfc (approved):
LGTM

deshpandevlab:
LGTM

PR #986, opened 2024-10-21
pgupgrade: add init container to delay startup of old deployment (#984) (PROJQUAY-8092)

bcaton85:
waits for existing postgresql pod to scale down before the old pod is up


deshpandevlab:
LGTM

deshpandevlab (approved):

ibazulic:
What about db containters where shutdown takes more than 30 seconds?

bcaton85:
this should cover most scenario's, @jonathankingfc is working on another PR that will ensure the previous deployment completes before starting the next which will go out in the next z-stream

PR #1002, opened 2024-11-07
fips:  Modified the postStart lifecycle hook to reference to sha256 (PROJQUAY-8185)

deshpandevlab:
### Testing

- manually verified PG upgrade

### DIFF
--- a/kustomize/components/clairpgupgrade/base/clair-pg-old.deployment.yaml
+++ b/kustomize/components/clairpgupgrade/base/clair-pg-old.deployment.yaml
@@ -65,10 +65,12 @@ spec:
                     echo 'Waiting for PostgreSQL to start...';
                     sleep 2;
                   done;
-                  echo 'password_encryption = scram-sha-256' >> $PGDATA/postgresql.conf;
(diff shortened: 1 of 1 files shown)


kustomize/components/clairpgupgrade/base/clair-pg-old.deployment.yaml
```diff
@@ -57,3 +57,20 @@ spec:
...
               memory: 2Gi
+          lifecycle:
+            postStart:
+              exec:
+                command: ["/bin/sh", "-c", "
+                  until pg_isready; do
+                    echo 'Waiting for PostgreSQL to start...';
+                    sleep 2;
+                  done;
+                  if [ \"\$(cat /proc/sys/crypto/fips_enabled)\" -eq 1 ]; then
```
bcaton85: you may want to change this to something like `if [[ -f /proc/sys/crypto/fips_enabled && $(< /proc/sys/crypto/fips_enabled) -eq 0 ]]; then` so it checks for the file

bcaton85: My linter says this results in "invalid escape sequence", i'm also getting an error saying it can't inflate this file

bcaton85 (approved):

PR #1038, opened 2025-05-08
cleanup: remove unused secrets (PROJQUAY-8900)

jonathankingfc:
## Summary by Sourcery

Improve cleanup of old configuration bundle secrets for Quay Registry

Bug Fixes:
- Fix potential issues with secret deletion by retrieving all matching old secrets

Enhancements:
- Add method to retrieve all old configuration bundle secrets
- Enhance secret cleanup process to handle multiple old secrets

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -448,6 +448,33 @@ func (r *QuayRegistryReconciler) GetConfigBundleSecret(
 	return bundle, nil
 }
 
+// GetOldConfigBundleSecrets returns all of the secrets with the prefix `<quayregistry-name>-config-bundle-`
+// in the same namespace as the QuayRegistry. This is used to clean up old config bundle secrets
+// when a new config bundle secret is created.
+func (r *QuayRegistryReconciler) GetOldConfigBundleSecrets(
+	ctx context.Context, quay *v1.QuayRegistry,
+
+) ([]corev1.Secret, error) {
+	secretList := &corev1.SecretList{}
+	if err := r.Client.List(
+		ctx,
+		secretList,
+		client.InNamespace(quay.GetNamespace()),
+	); err != nil {
+		return nil, err
+	}
+	if len(secretList.Items) == 0 {
+		return nil, goerrors.New("no unused secrets found")
+	}
+	var oldConfigBundleSecrets []corev1.Secret
+	for _, secret := range secretList.Items {
+		if strings.HasPrefix(secret.Name, quay.GetName()+"-quay-config-secret") && secret.Name != quay.Spec.ConfigBundleSecret {
+			oldConfigBundleSecrets = append(oldConfigBundleSecrets, secret)
+		}
+	}
+	return oldConfigBundleSecrets, nil
+}
+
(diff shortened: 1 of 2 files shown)


controllers/quay/quayregistry_controller.go
```diff
@@ -1055,30 +1079,38 @@ func (r *QuayRegistryReconciler) updateWithCondition(
...
+				ctx,
+				quay,
+				v1.ConditionTypeRolloutBlocked,
+				metav1.ConditionTrue,
+				v1.ConditionReasonComponentCreationFailed,
+				fmt.Sprintf("could not delete previous config bundle secret: %s", err),
 			)
 		}
-		log.Info("previous secret not found, continuing")
+		log.Info("successfully deleted previous config bundle secret")
```
bcaton85: this will print the same line for each secret, do we want to print the name of the secret to make this log more useful?

controllers/quay/quayregistry_controller.go
```diff
@@ -1055,30 +1079,38 @@ func (r *QuayRegistryReconciler) updateWithCondition(
...
-			Name:      quay.Spec.ConfigBundleSecret,
-			Namespace: quay.GetNamespace(),
-		},
-		previousSecret,
-	); err != nil {
-		r.Log.Info(fmt.Sprintf("%s secret not found, skipping", quay.Spec.ConfigBundleSecret))
-	}
-	if err := r.Client.Delete(ctx, previousSecret); err != nil {
+
+	log.Info("getting current config bundle secret")
```
bcaton85: is this log correct? it seems like it's just collecting and deleting all the existing config secrets

bcaton85 (changes requested):
Does this work correctly? It seems like it's still deleting all the existing config secrets, how is this preventing the deletion of the config secret that's currently being used?

controllers/quay/quayregistry_controller.go
```diff
@@ -448,6 +448,30 @@ func (r *QuayRegistryReconciler) GetConfigBundleSecret(
...
+	if err := r.Client.List(
+		ctx,
+		secretList,
+		client.InNamespace(quay.GetNamespace()),
+	); err != nil {
+		return nil, err
+	}
+	var oldConfigBundleSecrets []corev1.Secret
+	for _, secret := range secretList.Items {
+		if strings.HasPrefix(secret.Name, quay.GetName()+"-quay-config-secret") && secret.Name != quay.Spec.ConfigBundleSecret {
```
bcaton85, line 471: i'm assuming it doesn't deleting the existing secret because of the check here?

jonathankingfc: Yes, this prevents it from deleting the current secret

controllers/quay/quayregistry_controller.go
bcaton85: not sure if this log adds anything useful here, since we're not actually getting the config bundle secret, just deleting the other's. this would be more helpful if it said something along the lines of `garbage collecting outdated config secrets`

jonathankingfc: nice catch, this is an incorrect log statement. let me update this

controllers/quay/quayregistry_controller.go
jonathankingfc: cleaned up the logging here

bcaton85 (approved):

PR #1059, opened 2025-06-27
operator: add ephemeral volume support for Clair (PROJQUAY-6684)

dmesser:
### Summary

This PR introduces an optional feature for the managed Clair component, allowing users to enable a generic ephemeral volume mount at `/tmp` inside the Clair container. This addresses [PROJQUAY-6684](https://issues.redhat.com/browse/PROJQUAY-6684) and provides a workaround for issues with `/tmp` storage in certain environments. 

### Key Features

- **New Override:**  
  Adds a `useEphemeralVolume` boolean override for the managed Clair component. When set to `true`, a generic ephemeral volume is mounted at `/tmp`.
- **Configurable Storage:**  
  Users can optionally override `storageClassName` and `volumeSize` (default: 10Gi, minimum: 1Gi) for the ephemeral volume, similar to existing database components.
- **Validation:**  
    - If `useEphemeralVolume` is not enabled, specifying `storageClassName` or `volumeSize` triggers a clear error in the Clair component status.
    - Volume size below 1Gi is rejected with a user-friendly error.
- **Status Reporting:**  
  PVC provisioning and error events for the ephemeral volume are surfaced in the Clair component’s status, consistent with how database PVCs are handled.
- **Refactoring:**  
  Centralizes the logic and string constants for identifying the Clair deployment, improving maintainability.
- **Tests:**
  Unit test and integration test coverage has been added. Also an existing kuttl tests has been fixed.

### Motivation

In environments with a heavy load and large image layers, Clair doesn't have to use the executing nodes' (often limited) local storage for unpacking individual image layers, but can rely on external storage, which can dynamically request higher capacity or faster storage. It's not a classic persistent volume because we only need this storage during the lifetime of the individual pod.

### Backwards Compatibility

The new override is optional and defaults to `false`, preserving existing behavior for all users not opting in.

---

## Summary by Sourcery

Add optional ephemeral volume support for Clair component to mount a PVC-backed ephemeral volume at /tmp with configurable storage class and size, including validation, status reporting, and related tests.

New Features:
- Introduce a `useEphemeralVolume` override to enable an ephemeral PVC volume at `/tmp` in the managed Clair component
- Allow configuring the ephemeral volume’s `storageClassName` and `volumeSize` with sensible defaults

Enhancements:
- Validate that `storageClassName` or `volumeSize` overrides require `useEphemeralVolume=true` and enforce a minimum 1Gi volume size
- Surface PVC provisioning status and failure events for the ephemeral `/tmp` volume in the Clair component’s readiness condition
- Refactor deployment suffix constants and consolidate Clair deployment identification logic

Documentation:
- Extend the CRD to include the `useEphemeralVolume` field

Tests:
- Add unit and integration tests for ephemeral volume injection and status checks
- Add e2e KUTTL tests to verify `/tmp` volume mount and configuration


jbpratt:
fixed in #1225

PR #1064, opened 2025-07-07
PROJQUAY-8703: Read Only Root Filesystem

jacobsee:
Add readOnlyRootFilesystem to deployments to comply with product security guidelines.

## Summary by Sourcery

Enhancements:
- Add securityContext.readOnlyRootFilesystem: true to the base Quay deployment, Clair, ClairPostgres, Postgres, Redis, Mirror component deployments, and the Quay upgrade job.

### DIFF
--- a/kustomize/base/quay.deployment.yaml
+++ b/kustomize/base/quay.deployment.yaml
@@ -110,6 +110,8 @@ spec:
             limits:
               cpu: 2000m
               memory: 8Gi
+          securityContext:
+            readOnlyRootFilesystem: true
           startupProbe:
             httpGet:
               path: /health/instance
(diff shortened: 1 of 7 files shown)


jacobsee:
There were several issues with this naive attempt at moving to a read-only filesystem:

Multiple deployments (such as quay-app and quay-mirror) fetch extra certificates and write them to the container's trust store and Python site-packages directories on startup (https://github.com/quay/quay/blob/master/conf/init/certs_install.sh). The absolute path used depends on the version of Python in the container which is determined at runtime, so moving them out to a volume mount in Kustomize would be difficult. Further, making the container filesystem read only but then adding in mitigations so that the root trust store is writable again feels like missing the point of the security measure in the first place.

Multiple postgres instances are failing to start with simply a "pg_ctl: failed to start server" error under a read only filesystem, despite mounting in the log directory, tmp directory, and passwd file so that they're writable again.

The operator itself appears to dump a redis cache to disk on shutdown, so it does not shut down cleanly on a read only filesytem. This could probably be mitigated with a mount, but in my opinion there should be a discussion about whether that is the answer, what the value of that redis dump to the container filesystem is, etc.

PR #1093, opened 2025-12-02
only trigger postgres upgrade on non-matching repository name (PROJQUAY-9863)

bcaton85:
## Summary by Sourcery

Adjust Postgres upgrade detection to rely on repository name comparison instead of full image reference, handling differing registries, tags, and digests.

Bug Fixes:
- Prevent unnecessary Postgres upgrades when the image repository is unchanged but the registry, tag, or digest differs.

Tests:
- Add unit tests for image name extraction and repository name comparison to validate the new Postgres upgrade detection logic.


Marcusk19 (approved):
lgtm

PR #1100, opened 2026-02-02
monitoring: fix Grafana dashboard variable substitution (PROJQUAY-9386)

Marcusk19:
The Quay Grafana dashboard was showing "No datapoints found" due to
invalid characters in dashboard variable queries. The OpenShift Console
monitoring-plugin does not substitute Grafana template variables of
type "query" the same way a real Grafana instance would.

This fix makes three changes:

1. Change template variable type from "query" to "custom" in the
   dashboard ConfigMap template. Custom type variables have explicit
   options rather than dynamically queried ones.

2. Add "current" field to template variables, which is required for
   Grafana to properly apply the selected variable value.

3. Replace $namespace and $service placeholders directly in all panel
   expressions during reconciliation. This is necessary because the
   OpenShift Console monitoring-plugin does not substitute custom type
   variables into PromQL expressions.

Also adds a unit test to validate the dashboard template structure.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

## Summary by Sourcery

Fix Grafana dashboard variable handling so Quay metrics display correctly in the OpenShift Console monitoring plugin.

Bug Fixes:
- Ensure Grafana namespace and service dashboard variables have valid custom-type configuration so variable substitution no longer breaks metric queries.
- Populate current values for namespace and service variables so Grafana correctly applies selected values in the dashboard panels.
- Inline namespace and service placeholders into panel expressions to work around missing variable substitution in the OpenShift Console monitoring plugin.

Tests:
- Add a unit test validating the dashboard ConfigMap JSON structure and variable configuration to prevent regressions in Grafana variable handling.

### DIFF
--- a/controllers/quay/quayregistry_controller.go
+++ b/controllers/quay/quayregistry_controller.go
@@ -57,15 +57,19 @@ const (
 	creationPollInterval = time.Second * 2
 	creationPollTimeout  = time.Second * 600
 
-	grafanaDashboardConfigMapNameSuffix = "grafana-dashboard-quay"
-	grafanaTitleJSONPath                = "title"
-	grafanaNamespaceFilterJSONPath      = "templating.list.1.options.0.value"
-	grafanaServiceFilterJSONPath        = "templating.list.2.options.0.value"
-	clusterMonitoringLabelKey           = "openshift.io/cluster-monitoring"
-	quayDashboardJSONKey                = "quay.json"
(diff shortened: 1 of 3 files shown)


Marcusk19:
Validated with an openshift cluster. Where resource metrics were once empty, now we see data:
quay-dashboard-with-data

jbpratt (approved):

PR #1105, opened 2026-02-16
security: add container security hardening with CRD override (PROJQUAY-10168)

Marcusk19:
## Summary
- Adds hardened `securityContext` defaults (DROP ALL + ADD NET_BIND_SERVICE, `allowPrivilegeEscalation: false`, `runAsNonRoot: true`) to all Quay-owned containers: `quay-app`, `quay-mirror`, `quay-mirror-init`, and `quay-app-upgrade`
- Adds a `SecurityContext` override field to the `Override` struct in the CRD, supported on `quay` and `mirror` components, allowing users to opt out of the hardened defaults
- Applies the override in middleware for both Deployments (containers + init containers) and Jobs

## Test plan
- [x] `make test` — all existing and new unit tests pass
- [x] `make fmt && make vet` — code passes formatting and vetting
- [ ] Manual: deploy with `SKIP_RESOURCE_REQUESTS=true make run` and verify `securityContext` on quay-app, quay-mirror, and upgrade job containers
- [ ] Manual: apply a QuayRegistry with `securityContext` override on quay component, verify override takes effect
- [ ] Manual: verify existing QuayRegistry objects get hardened securityContext on next reconciliation


apis/quay/v1/quayregistry_types.go
```diff
@@ -521,8 +527,9 @@ func ValidateOverrides(quay *QuayRegistry) error {
 		hasstorageclass := component.Overrides.StorageClassName != nil
 		hasreplicas := component.Overrides.Replicas != nil
 		hasresources := component.Overrides.Resources != nil
+		hassecuritycontext := component.Overrides.SecurityContext != nil
 		hasenvvar := len(component.Overrides.Env) > 0
-		hasoverride := hasaffinity || hasvolume || hasenvvar || hasreplicas
+		hasoverride := hasaffinity || hasvolume || hasenvvar || hasreplicas || hassecuritycontext
```
jbpratt, line 532: Not blocking (existing behavior): Curious as to why we don't use `hasresources` and `hasstorageclass` in this conditional 🤔

pkg/middleware/middleware.go
```diff
@@ -180,6 +180,15 @@ func Process(quay *v1.QuayRegistry, qctx *quaycontext.QuayRegistryContext, obj c
 			ref.Resources.Limits = oresources.Limits
 		}
 
+		if osecctx := v1.GetSecurityContextOverrideForComponent(quay, kind); osecctx != nil {
```
jbpratt, line 183: Not blocking: perhaps these should merge with existing `securityContext` to preserve both (runUser from user provided, capabilities from default, etc)

jbpratt (approved):
Two non blocking comments!

PR #1112, opened 2026-03-04
PROJQUAY-10707: feat(tls): support external TLS secret reference for certificate rotation

### LINKED ISSUE #941: Allow the use of Cert-Manager to create certs for quay to use
I can't find an easy way to use Cert-Manager as the secrets that it creates are of type kubernetes.io/tls and have keys of tls.key and tls.cert.  Currently, the TLS config in the operator is looking for ssl.key and ssl.cert in the config-bundle-secret.  Can this be altered or added to so that tls.key and tls.cert are supported?

jbpratt:
## Summary

Add `secretRef` field to the Component struct enabling direct reference
to a `kubernetes.io/tls` Secret for HTTPS certificates when the TLS
component is unmanaged. The operator watches the referenced Secret and
triggers rolling restarts when its contents change, supporting automatic
certificate rotation with cert-manager or any Secret provider.

```yaml
spec:
  components:
  - kind: tls
    managed: false
    secretRef:
      name: my-tls-cert
```

- Add SecretRef to Component with CRD validation rules
- Implement checkExternalTLSSecret with conflict detection against
  config bundle ssl.cert/ssl.key
- Watch external TLS Secrets with data-change predicate (scoped via
  auto-labeling and label-filtered informer cache)
- Scope GenerationChangedPredicate to For/Owns sources only so Secret
  watches are not filtered by generation changes
- Add TLSSecretHash annotation to deployments for rolling restart
- Add ComponentTLSReady status reporting for secretRef mode
- Consolidate Secret type validation into pkg/tls.FetchAndValidate
- Add RBAC for Secret get/list/watch/patch

## Test plan

- [x] Unit tests for checkExternalTLSSecret (valid secret, missing secret,
  missing tls.crt/tls.key, ssl.cert/ssl.key conflicts, label persistence)
- [x] Unit tests for ComponentTLSReady status evaluator
- [x] Unit tests for findQuayRegistriesForSecret mapper
- [x] Unit tests for hasNecessaryConfig TLS skip
- [x] Dedicated `external_tls` chainsaw e2e test:
  - Fresh deploy with secretRef from the start
  - Certificate rotation (new cert, hash change, rolling restart)
  - Secret deletion and self-healing (ComponentTLSReady=False → recreate → recovery)
- [x] KinD CI passing (21 min)

closes #941


jbpratt:
## Manual Test Results: External TLS Secret with cert-manager Certificate Rotation

**Environment:** ROSA cluster, operator via `SKIP_RESOURCE_REQUESTS=true make run` (local), cert-manager v1.19.4

### Setup

1. Installed cert-manager:
   ```
   kubectl apply -f https://github.com/cert-manager/cert-manager/releases/latest/download/cert-manager.yaml
   ```

2. Installed CRDs and ran operator locally:
   ```
   make install
   SKIP_RESOURCE_REQUESTS=true make run
   ```

3. Created test namespace:
   ```
   kubectl create namespace quay-tls-test
   ```

4. Created a self-signed `ClusterIssuer`:
   ```yaml
   apiVersion: cert-manager.io/v1
   kind: ClusterIssuer
   metadata:
     name: selfsigned-issuer
   spec:
     selfSigned: {}
   ```

5. Created a `Certificate` CR with the correct route SAN:
   ```yaml
   apiVersion: cert-manager.io/v1
   kind: Certificate
   metadata:
     name: quay-tls
   spec:
     secretName: quay-tls
     duration: 2160h
     renewBefore: 360h
     issuerRef:
       name: selfsigned-issuer
       kind: ClusterIssuer
     dnsNames:
       - <quay-route-hostname>
   ```

6. Created `QuayRegistry` with external TLS secretRef:
   ```yaml
   kind: QuayRegistry
   metadata:
     name: quay-tls-test
   spec:
     components:
       - kind: tls
         managed: false
         secretRef:
           name: quay-tls
       - kind: monitoring
         managed: false
       - kind: mirror
         managed: false
       - kind: clair
         managed: false
       - kind: clairpostgres
         managed: false
       - kind: horizontalpodautoscaler
         managed: false
   ```

NooBaa was pre-configured for object storage.

---

### Test 1: Initial Deployment with External TLS Secret — PASS

**Steps:**
1. Applied the ClusterIssuer, Certificate, and QuayRegistry resources above
2. Waited for cert-manager to populate the `quay-tls` Secret
3. Waited for operator reconciliation and all pods to become ready

**Results:**
- `ComponentTLSReady=True` with message "Using externally managed TLS certificate"
- `Available=True` / "All components reporting as healthy"
- `quay-operator/tls-secret-hash` annotation (`0e42350f`) present on all deployments (quay-app, quay-database, quay-redis) and their pod templates
- x509 SAN validation correctly rejected a cert with wrong hostname (`RolloutBlocked=True` with "certificate is valid for X, not Y"), then succeeded after correcting the Certificate CR's `dnsNames`

---

### Test 2: Certificate Rotation via `kubectl cert-manager renew` — PASS

**Steps:**
1. Recorded current hash: `c6f598c7`
2. Forced renewal:
   ```
   kubectl cert-manager renew quay-tls -n quay-tls-test
   ```
3. Waited for certificate issuance and operator reconciliation
4. Verified hash and rollout status

**Results:**
- Hash changed: `c6f598c7` → `9ab2371e`
- Rolling restart observed (old pods terminated, new pods started with updated hash)
- `ComponentTLSReady=True` and `Available=True` after rollout completed

---

### Test 3: Secret Deletion + Auto-Recovery — PASS

**Steps:**
1. Deleted the TLS Secret:
   ```
   kubectl delete secret quay-tls -n quay-tls-test
   ```
2. Waited for cert-manager to recreate the Secret (Certificate CR still exists)
3. Checked operator conditions and hash

**Results:**
- cert-manager recreated the Secret within seconds
- Hash changed on recreation, rolling restart occurred
- Operator recovered automatically without manual intervention
- `ComponentTLSReady=True` and `Available=True` after recovery
- The `cmpstatus.TLS.Check()` path correctly returns "External TLS secret not found" when the Secret is missing (transient state was brief due to cert-manager's fast recreation)

---

### Test 4: Manual Secret Update (non-cert-manager) — PASS

**Steps:**
1. Deleted the Certificate CR to prevent cert-manager from overwriting:
   ```
   kubectl delete certificate quay-tls -n quay-tls-test
   ```
2. Recorded current hash: `863cb9b4`
3. Generated a new self-signed cert and applied it:
   ```
   openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
     -keyout /tmp/new.key -out /tmp/new.crt \
     -subj "/CN=manual-test" \
     -addext "subjectAltName=DNS:<quay-route-hostname>"

   kubectl create secret tls quay-tls \
     --cert=/tmp/new.crt --key=/tmp/new.key \
     --dry-run=client -o yaml | kubectl apply -f - -n quay-tls-test
   ```
4. Waited for operator reconciliation and rollout

**Results:**
- Hash changed: `863cb9b4` → `3b6a71f5`
- Rolling restart occurred, Quay came back healthy
- `ComponentTLSReady=True` and `Available=True`

---

### Hash Progression Across All Tests

| Event | Hash |
|-------|------|
| Initial cert-manager cert | `0e42350f` |
| Secret delete + cert-manager recreate | `af0bbc6f` |
| Second recreation | `c6f598c7` |
| `kubectl cert-manager renew` | `9ab2371e` |
| cert-manager auto-renewal | `863cb9b4` |
| Manual openssl cert | `3b6a71f5` |

### Verification Checklist

- [x] Initial deployment succeeds with cert-manager Secret
- [x] cert-manager renewal triggers rolling restart with new hash
- [x] Secret deletion causes error, automatic recovery on recreation
- [x] Manual Secret update triggers rotation
- [x] `ComponentTLSReady` condition correctly reports external TLS status
- [x] x509 SAN validation rejects mismatched certs
- [x] Hash annotation on both deployment metadata and pod template
- [x] All managed deployments (quay-app, quay-database, quay-redis) receive the hash
- [x] Secret watch predicate filters on data changes only

pkg/tls/validate.go
```diff
@@ -0,0 +1,35 @@
...
+
+	cert, certOK := s.Data["tls.crt"]
+	key, keyOK := s.Data["tls.key"]
+	if !certOK || len(cert) == 0 {
+		return nil, nil, nil, fmt.Errorf("TLS secret %q missing or empty tls.crt", name)
+	}
+	if !keyOK || len(key) == 0 {
+		return nil, nil, nil, fmt.Errorf("TLS secret %q missing or empty tls.key", name)
+	}
+
```
tlwu2013, line 37: The current `FetchAndValidate` checks that `tls.crt` and `tls.key` are non-empty, but doesn't validate the content. 

This means an **expired** cert, a cert/key **mismatch**, or **corrupted** PEM would all pass through silently, then the operator would annotate the deployment, trigger a rolling restart, and finally Quay pods would fail to start (or start serving an expired cert).

I'd suggest adding three lightweight checks using Go stdlib only (`crypto/tls`, `crypto/x509`, `encoding/pem` — no new deps):
```
// Validates PEM format and that the private key matches the certificate.
if _, err := tls.X509KeyPair(cert, key); err != nil {
    return nil, nil, nil, fmt.Errorf("TLS secret %q contains an invalid key pair: %w", name, err)
}

// Validate the certificate is not already expired.
block, _ := pem.Decode(cert)
if block == nil {
    return nil, nil, nil, fmt.Errorf("TLS secret %q: tls.crt is not valid PEM", name)
}
x509Cert, err := x509.ParseCertificate(block.Bytes)
if err != nil {
    return nil, nil, nil, fmt.Errorf("TLS secret %q: failed to parse certificate: %w", name, err)
}
if time.Now().After(x509Cert.NotAfter) {
    return nil, nil, nil, fmt.Errorf("TLS secret %q: certificate expired at %s", name, x509Cert.NotAfter.UTC().Format(time.RFC3339))
}
```
This way:
- `tls.X509KeyPair` handles PEM parsing, key format, and cert/key match in one call. 
- The expiry check prevents the operator from applying a cert that is already past its `NotAfter`  (the error surfaces as a `ConditionTypeRolloutBlocked` condition), so the user gets actionable feedback rather than a Quay pod crash loop.

And the existing unit tests in `features_test.go` for malformed secrets would also need to be extended with corresponding cases (**_i.e.,_** invalid PEM, mismatched key, expired cert) once these checks are added.

jbpratt: Hey @tlwu2013 , it looks like we already support 2 of the 3 mentioned checks in config-tool (PEM format, cert parsing, as well as validating hostname matches SAN/CN). The operator calls config-tool to validate the cert already for all certificate paths https://github.com/quay/quay-operator/blob/dfd7ab42ad3df61b36157a91df128d16262c2fc9/pkg/kustomize/secrets.go#L211 

It does not do an expiry check, but I think that would be best to go into config-tool so it runs on all cert validation, not just external secret ref. What do you think?

tlwu2013: I see.  I agree that adding the expiry check to `ValidateCertPairWithHostname` in the config-tool is the right place, so all cert paths benefit.  Will you be helping with a separate PR to the config-tool for this?

With that said, if we agree to defer the validation check outside of `FetchAndValidate`, do you think it's worth adding a comment in `FetchAndValidate` to call out that this function only checks the presence, and the deeper validation would happen later in the call flow via the config-tool?

jbpratt: Please open up a ticket for the config-tool check and we can prioritize it!

tlwu2013: Got it.  I've looked up the config-tool repo and figured it's now read-only due to being archived. 

I'd say let's defer the config-tool change then.  Great work @jbpratt!

main.go
```diff
@@ -85,7 +86,19 @@ func main() {
...
+	// so uncached reads provide read-after-write consistency.
+	cacheOptions := cache.Options{
+		ByObject: map[client.Object]cache.ByObject{
+			&corev1.Secret{}: {
+				Label: labels.SelectorFromSet(labels.Set{
+					quay.TLSSecretLabel: "true",
+				}),
+			},
+		},
+	}
```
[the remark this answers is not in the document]
jbpratt:
```
Initial provisioning is fine — the QuayRegistry creation triggers reconcile (via For(&v1.QuayRegistry{})), which fetches the Secret with an uncached read, adds the label, and the watch bootstraps. The Secret watch is only needed
   for subsequent changes.

  Delete/recreate is handled by the 10-second requeue:
  - reconcileWithCondition returns r.Requeue = ctrl.Result{RequeueAfter: 10 * time.Second} (main.go:151)
  - So when checkExternalTLSSecret fails (Secret gone), the controller requeues every 10s
  - Within 10s of the Secret being recreated, the next requeue finds it, labels it, and the watch resumes

  Caching all Secrets (their suggestion) would:
  - Watch every Secret in the namespace — potentially thousands on busy clusters
  - Generate event noise from unrelated Secrets (even if filtered by the predicate/mapper, events still enter the queue)
  - Use significantly more memory

  The label-filtered informer is a deliberate optimization. The 10s requeue gap on delete/recreate is perfectly acceptable
```

Marcusk19 (approved):

PR #1221, opened 2026-03-25
PROJQUAY-6578: fix(reconciler): defer workloads until managed databases are ready

jbpratt:
The migration job could start before postgres was available, causing
partial migrations and DuplicateTable errors on retry. Defer dependent
Deployments and Jobs until managed database deployments have available
replicas, while always creating infrastructure deployments (postgres,
clair-postgres, redis).

### DIFF
--- a/kustomize/components/clair/clair.deployment.yaml
+++ b/kustomize/components/clair/clair.deployment.yaml
@@ -29,7 +29,7 @@ spec:
                         - clair-app
                 topologyKey: "kubernetes.io/hostname"
       containers:
-        - image: quay.io/projectquay/clair:nightly
+        - image: quay.io/projectquay/clair:4.8.0
           imagePullPolicy: IfNotPresent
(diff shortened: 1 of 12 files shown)


kustomize/components/clair/clair.deployment.yaml
```diff
@@ -29,7 +29,7 @@ spec:
                         - clair-app
                 topologyKey: "kubernetes.io/hostname"
       containers:
-        - image: quay.io/projectquay/clair:nightly
+        - image: quay.io/projectquay/clair:4.8.0
```
Marcusk19, line 32: should this be 4.9.0?

jbpratt: Yes but this isn't actually used - we override it with the environment variable. I'll follow up to clean this up!

Marcusk19 (approved):

PR #1222, opened 2026-03-27
PROJQUAY-10705: feat(tls): inherit cluster TLS security profile for Quay and Mirror 

Marcusk19:
## Summary

- Read the cluster-wide `tlsSecurityProfile` from the OpenShift `APIServer` resource and translate it to Quay's `SSL_PROTOCOLS`/`SSL_CIPHERS` config format
- Inject inherited TLS settings into the generated `config.yaml` during `Inflate()`
- User overrides via `configBundleSecret` take precedence — if `SSL_PROTOCOLS` or `SSL_CIPHERS` are already set, the cluster profile is not applied
- Vanilla Kubernetes (no `config.openshift.io` API) falls back gracefully with no TLS settings injected
- Mirror workers inherit automatically via the shared config

## Changes

| File | Purpose |
|------|---------|
| `controllers/quay/tls.go` | TLS profile detection (`checkTLSSecurityProfile`) and translation logic |
| `controllers/quay/tls_test.go` | Unit + integration tests (15 test cases) |
| `controllers/quay/quayregistry_controller.go` | Call site in reconcile loop + RBAC marker |
| `pkg/context/context.go` | `SSLProtocols` and `SSLCiphers` fields |
| `pkg/kustomize/kustomize.go` | Inject TLS settings into generated config |
| `main.go` | Register `configv1` scheme |
| `config/rbac/role.yaml` | `get`/`watch` on `config.openshift.io/apiservers` |
| `docs/tls-security-profile.md` | Feature documentation and e2e testing guide |
| `vendor/` | Vendor `github.com/openshift/api/config/v1` package |

## Test plan

- [x] Unit tests pass: `go test ./controllers/quay/ -run 'TestTranslateTLS|TestCheckTLS|TestTls' -v` (15/15 passing)
- [x] E2E validated on CRC v4.21.0:
  - [x] Nil cluster profile → Intermediate defaults (`SSL_PROTOCOLS: [TLSv1.2, TLSv1.3]`, 11 ciphers)
  - [x] Modern profile propagation (`SSL_PROTOCOLS: [TLSv1.3]`, 3 TLS 1.3 ciphers)
  - [x] User override in `configBundleSecret` takes precedence over cluster profile
- [x] Vanilla K8s graceful fallback (tested via fake client without configv1 scheme)

## Jira

[PROJQUAY-10705](https://issues.redhat.com/browse/PROJQUAY-10705)


pkg/kustomize/kustomize.go
```diff
@@ -650,6 +650,20 @@ func Inflate(
...
+		if _, ok := parsedUserConfig["SSL_PROTOCOLS"]; !ok {
+			parsedUserConfig["SSL_PROTOCOLS"] = strings.Split(ctx.SSLProtocols, " ")
+		}
+	}
+	if ctx.SSLCiphers != "" {
+		if _, ok := parsedUserConfig["SSL_CIPHERS"]; !ok {
+			parsedUserConfig["SSL_CIPHERS"] = strings.Split(ctx.SSLCiphers, ":")
+		}
+	}
+
```
ibazulic, line 680: These only make sense if TLS is terminated on Quay. If TLS is terminated on the cluster, which is the default value, then injecting these will do nothing, between haproxy and Quay communication is unencrypted. So these should be applied only if route/TLS are unmanaged components.

controllers/quay/quayregistry_controller.go
```diff
@@ -1235,6 +1248,11 @@ func (r *QuayRegistryReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 		return err
 	}
+	if err := configv1.Install(mgr.GetScheme()); err != nil {
+		r.Log.Error(err, "Failed to add OpenShift `config` API to scheme")
+
```
jbpratt:
```suggestion
```

formatting

jbpratt (changes requested):
quay integration looks good to me, let's confirm if we need to propagate this to all our managed resources (postgres, redis, clair) as well as add a chainsaw openshift specific test please!

pkg/kustomize/kustomize.go
Marcusk19: Thanks, added another check to make sure routes are unmanaged before injecting

jbpratt (approved):

controllers/quay/quayregistry_controller.go
```diff
@@ -486,6 +487,7 @@ func (r *QuayRegistryReconciler) GetOldConfigBundleSecrets(
 // +kubebuilder:rbac:groups=objectbucket.io,resources=objectbucketclaims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules;servicemonitors,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=config.openshift.io,resources=apiservers,verbs=get;watch
```
jbpratt:
```suggestion
// +kubebuilder:rbac:groups=config.openshift.io,resources=apiservers,verbs=get;list;watch
```

I think we need list for watching with the informer

jbpratt: actually nevermind, let's try this fix
```
  |     Fix (option B): Add `&configv1.APIServer{}` to the                       |
  |       `client.CacheOptions.DisableFor` list in main.go so the client         |
  |       bypasses the cache and makes a direct API call (only needs `get`).     |
  |       This is actually more appropriate since you only need a single         |
  |       Get, not a continuous watch/cache.                                     |
```

jbpratt: we should be able to drop watch from the verbs if we do this

jbpratt (approved):
It should auto merge when the prow job finishes

PR #1225, opened 2026-03-30
PROJQUAY-11162: fix(clair): mount volume to /var/tmp

jbpratt:
Overlayfs doesn't support O_TMPFILE so clair was filling up writable storage

### DIFF
--- a/kustomize/components/clair/clair.deployment.yaml
+++ b/kustomize/components/clair/clair.deployment.yaml
@@ -71,6 +71,8 @@ spec:
               subPath: 01_user_config.yaml
             - mountPath: /var/run/certs
               name: certificates
+            - name: indexer-layer-storage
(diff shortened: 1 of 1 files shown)


ibazulic:
We need to add a PVC definition as well to the templates so that the operator can create it and then mount it. Currently, only the change in the deployment file exists.

jbpratt:
> We need to add a PVC definition as well to the templates so that the operator can create it and then mount it. Currently, only the change in the deployment file exists.

@ibazulic this is using ephemeral volumes https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes

ibazulic (approved):

Marcusk19 (approved):

PR #1226, opened 2026-03-30
PROJQUAY-11120: fix(clair): update updaters config for clair/config v1.4.3

Marcusk19:
## Summary
- Rename `rhel` updater to `rhel-vex` in generated Clair config to match clair/config v1.4.3 naming
- Add explicit `sets` list excluding `osv` to prevent all updaters activating by default, which caused "Invalid Semantic Version" errors blocking vulnerability database pulls
- Add `TestClairConfigUpdaters` to validate the generated updater config

## Test plan
- [x] `go test ./pkg/kustomize/...` passes
- [ ] Deploy on cluster and verify Clair can pull vulnerability database without errors
- [ ] Confirm `osv` updater does not activate unless explicitly configured

### DIFF
--- a/pkg/kustomize/secrets.go
+++ b/pkg/kustomize/secrets.go
@@ -448,8 +448,20 @@ func clairConfigFor(log logr.Logger, qctx *quaycontext.QuayRegistryContext, quay
 			"name": "prometheus",
 		},
 		"updaters": map[string]interface{}{
+			"sets": []string{
+				"alpine",
+				"aws",
+				"clair.cvss",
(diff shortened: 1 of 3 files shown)


jbpratt (approved):
should auto merge when openshift e2e passes

jbpratt:
/cherrypick redhat-3.17 

will need a manual cherrypick to 3.16 when we bump. I'll try to cherrypick the whole chainsaw suite back to 3.16 soon so these will work again

PR #1237, opened 2026-04-03
PROJQUAY-11212: feat(route): support annotation and label overrides on managed routes

### LINKED ISSUE #310: External access Openshift route annotation
I'm using the Quay operator on Openshift and somehow need to limit the access to the Quay and Quay config routes to a list of IP addresses.
I tried to add the annotations to `spec.quay.externalAcccess.annotations` but it seems like this field is only relevant when the `externalAccess.type` is `ingress`.

Example:
```
[...]
    externalAccess:
      hostname: quay.apps.<domain>
      configHostname: quay-config.apps.<domain>
      annotations:
        haproxy.router.openshift.io/ip_whitelist: x.x.x.x,y.y.y.y
      tls:
        secretName: quay-letsencrypt-cert
        termination: passthrough
[...]
```
When I add annotations to the routes manually, the operator reverts that change at some point.

Is there a way to add annotations to the created Openshift routes?

jbpratt:
Apply user-specified annotations and labels from the route component's
Override to managed OpenShift Route objects. The Override struct and
getter functions already existed but were never wired up for Routes
in the middleware layer.

Closes: https://github.com/quay/quay-operator/issues/310

### DIFF
--- a/pkg/middleware/middleware.go
+++ b/pkg/middleware/middleware.go
@@ -301,6 +301,27 @@ func Process(quay *v1.QuayRegistry, qctx *quaycontext.QuayRegistryContext, obj c
 		return obj, nil
 	}
 
+	if olabels := v1.GetLabelsOverrideForComponent(quay, v1.ComponentRoute); olabels != nil {
+		if rt.Labels == nil {
+			rt.Labels = map[string]string{}
+		}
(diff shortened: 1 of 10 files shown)


shaonrh (approved):
LGTM

PR #1277, opened 2026-06-10
PROJQUAY-11215: feat: add TLS support for operator-managed PostgreSQL

bcaton85:
Adds opt-in TLS encryption for operator-managed PostgreSQL instances used by Quay Registry and Clair, enabling customers in regulated industries to meet encryption-in-transit requirements without provisioning external database infrastructure.

  What this enables

  spec:
    components:
    - kind: postgres
      managed: true
      overrides:
        tls:
          enabled: true
    - kind: clairpostgres
      managed: true
      overrides:
        tls:
          enabled: true

  - PostgreSQL pods are configured with ssl = on and serve TLS connections
  - Quay and Clair connection strings are updated with sslmode=verify-full
  - Self-signed certificates (ECDSA P-256, 10yr) are generated automatically, or users can provide their own via secretRef (compatible with cert-manager)
  - Existing deployments are unaffected — TLS is only enabled when explicitly configured

  Certificate options

  - Self-signed (tls.enabled: true) — Operator generates ECDSA P-256 certs automatically. Best for quick start and dev/test.
  - User-provided (tls.secretRef.name: my-certs) — Reference a Secret containing ca.crt, tls.crt, tls.key. For enterprise PKI integration.
  - cert-manager (tls.secretRef.name: cert-manager-secret) — Point secretRef at a cert-manager-populated Secret. Operator watches for changes and triggers reconciliation on
  rotation.

  Commits
  
  1. apis: TLSOverride struct on the Override type, CEL validation, restricted to postgres/clairpostgres
  2. kustomize (certs): ECDSA P-256 cert generation, persisted in managed keys Secret
  3. kustomize (connection strings): sslmode=verify-full on Quay DB_URI and Clair connstring, CA Secret generation for both self-signed and secretRef modes
  4. middleware: Projected volume with defaultMode: 0600, init container for postgresql.conf patching (existing PVCs), ConfigMap patching (fresh PVCs), cleanup init container for downgrade
  5. controller: secretRef validation (key presence, cert/key match, expiry), RolloutBlocked on failure, Secret watch for cert-manager rotation
  6. e2e tests: 4 Chainsaw test suites

  Test plan

  Unit tests

  - CRD validation: TLS override accepted on postgres/clairpostgres, rejected on unsupported components
  - Certificate generation: valid chain, correct SANs, PEM encoding, ECDSA P-256
  - Connection strings: DB_URI and Clair connstring with/without TLS, existing params preserved
  - Middleware: init container injection, volume mounts, ConfigMap patching, cleanup on disable
  - Validation: missing secret, missing keys, cert/key mismatch, expired cert

  E2E tests (Chainsaw, all passing)

  - [x] postgres_tls — self-signed TLS on both components, cert SANs/chain verification, SHOW ssl = on
  - [x] postgres_tls_mixed — TLS on postgres only, clairpostgres uses sslmode=disable
  - [x] postgres_tls_secretref — user-provided certs work; invalid secretRef sets RolloutBlocked
  - [x] postgres_tls_lifecycle — enable TLS on existing deployment, verify SSL active, disable TLS, verify clean downgrade

  Manual verification on OpenShift

  - [x] Deploy with TLS, create user, create org, push image — all over encrypted DB connections
  - [x] Disable TLS, verify clean downgrade, create org — works without TLS
  - [x] Re-enable TLS, verify SSL back on, create org — works with TLS restored
  - [x] SHOW ssl returns on/off correctly in each state
  - [x] pg_stat_ssl confirms encrypted remote connections from Clair

  Enhancement: https://github.com/quay/enhancements/pull/41


LiZhang19817:
@bcaton85 pls review this critical bug when using unmanaged postgres database with TLS Cert.

```
  Critical Bug
  
  1. DB_URI sslmode stripped from unmanaged postgres connections

  pkg/kustomize/kustomize.go — the new else branch at the TLS check:

  if override := v1.GetTLSOverrideForComponent(quay, v1.ComponentPostgres); override != nil && override.Enabled {
      // add sslmode=verify-full ...
  } else {
      // STRIPS sslmode and sslrootcert from ALL DB_URIs
      q.Del("sslmode")
      q.Del("sslrootcert")
  }

  This runs unconditionally — including for unmanaged postgres where the user provides their own DB_URI with sslmode=verify-full to
  connect to an external TLS-enabled PostgreSQL. Since overrides can't be set on unmanaged components, GetTLSOverrideForComponent
  always returns nil, and the else branch strips the user's TLS params.
  
  Impact: Any existing deployment with unmanaged postgres using TLS connections to external databases would have sslmode and
  sslrootcert silently removed from their DB_URI, breaking database connections on the first reconcile after operator upgrade.

  Fix: Guard the else branch with a ComponentIsManaged check:
  } else if v1.ComponentIsManaged(quay.Spec.Components, v1.ComponentPostgres) {
      // only strip TLS params from managed postgres DB_URI
```

pkg/kustomize/certs.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	caSerial, err := randomSerial()
+	if err != nil {
+		return "", "", "", err
+	}
+
+	now := time.Now()
+	caTemplate := &x509.Certificate{
+		SerialNumber: caSerial,
```
jbpratt: you can remove this randomSerial bit and setting this in the Certificate - `CreateCertificate` does it at runtime (coderabbit and claude get this wrong often)

pkg/kustomize/certs.go
```diff
@@ -0,0 +1,142 @@
...
+		return "", "", "", fmt.Errorf("generating server key: %w", err)
+	}
+
+	serverSerial, err := randomSerial()
+	if err != nil {
+		return "", "", "", err
+	}
+
+	serverTemplate := &x509.Certificate{
+		SerialNumber: serverSerial,
```
jbpratt: same as above

pkg/kustomize/kustomize.go
```diff
@@ -674,6 +785,46 @@ func Inflate(
 	}
 	parsedUserConfig["DB_URI"] = ctx.DbUri
 
+	if ctx.PostgresUseServiceCA {
+		u, err := url.Parse(ctx.DbUri)
+		if err == nil {
+			shortHost := quay.GetName() + "-quay-database"
+			fqdn := shortHost + "." + quay.GetNamespace() + ".svc"
+			if u.Hostname() == shortHost {
+				u.Host = fqdn + ":" + u.Port()
```
jbpratt: **nit: trailing colon when user-provided DB_URI omits port**

`u.Port()` returns `""` when the URI has no explicit port, which produces `host.ns.svc:` (trailing colon). The auto-generated path always includes `:5432` so this only affects user-provided `DB_URI` in the config bundle — but it could silently break postgres connections for those users.

```go
// safer:
port := u.Port()
if port == "" {
    port = "5432"
}
u.Host = fqdn + ":" + port
```

pkg/kustomize/kustomize.go
```diff
@@ -420,6 +421,42 @@ func KustomizationFor(
 		addrs = append(addrs, noproxy...)
 	}
 
+	if v1.ComponentIsManaged(quay.Spec.Components, v1.ComponentPostgres) {
+		if override := v1.GetTLSOverrideForComponent(quay, v1.ComponentPostgres); override != nil && override.Enabled && override.SecretRef == nil {
+			if ctx.PostgresUseServiceCA {
+				log.Info("using OpenShift service CA for managed postgres TLS, skipping self-signed cert generation")
+			} else if ctx.PostgresTLSCA == "" {
+				log.Info("generating self-signed TLS certificates for managed postgres")
+				serviceName := quay.GetName() + "-quay-database"
```
jbpratt, line 430: **nit: service name constructed inline — consider reusing `postgresServiceName()`**

This name `{name}-quay-database` is constructed inline here and at lines 448, 775, 791 in this file, plus `secrets.go:407`. Meanwhile `features.go` already centralizes this in `postgresServiceName()`. If one of these ever drifts (e.g., renamed in Kustomize base), the cert SANs or connection strings would silently break.

Not blocking — but worth extracting into a shared constant or moving `postgresServiceName()` to a common package so both `kustomize` and `controllers` can use it.

pkg/kustomize/certs.go
bcaton85: done

pkg/kustomize/kustomize.go
bcaton85: done

pkg/kustomize/kustomize.go
bcaton85: this ends up being a larger change that would probably be better as a follow up. Reusing that function creates a circular import so it needs to be externalized as a separate package and will require sprawling changes. Let's leave this one be for now

jbpratt (approved):

PR #1279, opened 2026-06-17
PROJQUAY-11851: fix(bundle): move apiservers RBAC to clusterPermissions in CSV

Marcusk19:
## Summary

The `apiservers.config.openshift.io` GET rule was under `permissions` (namespace-scoped) in the CSV instead of `clusterPermissions` (cluster-scoped). OLM translates `permissions` into a namespace-scoped Role, which cannot grant access to cluster-scoped resources — Kubernetes silently ignores the rule. This caused a Forbidden error on every OLM install, blocking QuayRegistry reconciliation in an infinite retry loop.

**Root cause:** CSV misconfiguration, not a code bug. The `config/rbac/role.yaml` used by `make run` is a ClusterRole and worked correctly, which is why it was never caught in dev.

**Fix:** Move the `apiservers` rule from `permissions` to a new `clusterPermissions` section in the CSV so OLM creates a ClusterRole + ClusterRoleBinding for it.

Fixes: [PROJQUAY-11851](https://issues.redhat.com/browse/PROJQUAY-11851)

## Test plan

- [x] `make test` — full test suite passes
- [x] OLM e2e on CRC: installed operator via CatalogSource → Subscription, verified OLM created ClusterRole with `apiservers` GET rule, `oc auth can-i get apiservers` returns `yes`, operator runs with zero Forbidden errors

### DIFF
--- a/controllers/quay/tls.go
+++ b/controllers/quay/tls.go
@@ -52,15 +52,6 @@ func (r *QuayRegistryReconciler) checkTLSSecurityProfile(
 		if meta.IsNoMatchError(err) || runtime.IsNotRegisteredError(err) {
 			return nil
 		}
-		// Forbidden means the APIServer resource exists but the operator lacks
-		// permission to read it (e.g. OLM did not apply the ClusterRole). Log
-		// and continue so that a missing RBAC binding does not block the entire
-		// QuayRegistry deployment — the operator falls back to Quay's built-in
-		// TLS defaults.
-		if errors.IsForbidden(err) {
-			r.Log.Info("unable to read cluster TLS security profile due to missing RBAC, falling back to default", "error", err)
-			return nil
-		}
 		return fmt.Errorf("fetching APIServer: %w", err)
 	}
 
--- a/controllers/quay/tls.go
+++ b/controllers/quay/tls.go
@@ -52,8 +52,13 @@ func (r *QuayRegistryReconciler) checkTLSSecurityProfile(
 		if meta.IsNoMatchError(err) || runtime.IsNotRegisteredError(err) {
 			return nil
 		}
+		// Forbidden means the APIServer resource exists but the operator lacks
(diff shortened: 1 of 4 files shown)


Marcusk19:
Re: fail-open / degraded condition suggestion — this is a valid observation but out of scope for this PR. Today there is no `Degraded` condition type in the QuayRegistry API — adding one would introduce a new contract on the CRD surface (monitoring, OLM, scripts all consume conditions). The info-level log is emitted on every reconciliation cycle so it remains visible in log aggregation. If a `Degraded` condition is desired, it should be tracked as a separate issue since it would benefit other non-blocking warnings beyond just TLS profile inheritance.

controllers/quay/tls.go
```diff
@@ -52,6 +52,15 @@ func (r *QuayRegistryReconciler) checkTLSSecurityProfile(
 		if meta.IsNoMatchError(err) || runtime.IsNotRegisteredError(err) {
 			return nil
 		}
+		// Forbidden means the APIServer resource exists but the operator lacks
+		// permission to read it (e.g. OLM did not apply the ClusterRole). Log
+		// and continue so that a missing RBAC binding does not block the entire
```
jbpratt: Does it ever apply the ClusterRole? Is this a race issue or a misconfiguration?

Marcusk19: I think this is hitting a gap in my understanding, I thought originally this was just an extra error handler we needed for user misconfiguration. but looking further I see that it appears there's a rule for the apiserver permission that's in the wrong place within the CSV. I guess this is what informs OLM to create the ClusterRole etc. and it was never applied? Submitted a patch to also fix the CSV entry.

jbpratt: We should be able to remove this and only do the CSV change

Marcusk19: done!

jbpratt (approved):

PR #1281, opened 2026-06-17
PROJQUAY-11867: fix(apis): restrict secretRef to tls component only

Rachimypala-RM:
secretRef was exposed on the generic Component struct without any kind-level guard, allowing it to be set on components such as postgres, redis, and clair where it has no effect. The only consumer of SecretRef in the codebase is GetTLSSecretRef, which already gates on kind == tls.

Add a supportsSecretRef allowlist (tls only), a new CEL admission rule that rejects secretRef on any component whose kind is not tls, and a ValidateOverrides runtime check that fires during reconciliation as a belt-and-suspenders guard.

## Release Notes

* **New Features**
  * TLS components now support `secretRef` overrides.

* **Bug Fixes**
  * Added validation to reject `secretRef` overrides on unsupported component types, including PostgreSQL, Redis, and Clair.
  * Validation now provides clear feedback when `secretRef` is used outside TLS components.

* **Tests**
  * Expanded coverage for supported and unsupported `secretRef` configurations.

### DIFF
--- a/apis/quay/v1/quayregistry_types.go
+++ b/apis/quay/v1/quayregistry_types.go
@@ -133,6 +133,10 @@ var supportsTLSOverride = []ComponentKind{
 	ComponentClairPostgres,
 }
 
+var supportsSecretRef = []ComponentKind{
+	ComponentTLS,
+}
+
 const (
 	ManagedKeysName             = "quay-registry-managed-secret-keys"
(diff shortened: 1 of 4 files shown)


apis/quay/v1/quayregistry_types.go
jbpratt, line 1: @Rachimypala-RM could we also get some unit tests to validate this fix? Thank you!

PR #1283, opened 2026-06-18
PROJQUAY-11892: fix: add scraping for quay-monitor pods

rgordill:
Now every pod created by the operator that should be monitored, is included in the service selector.


jbpratt:
Superseded by #1286, which includes all changes from this PR plus protection for the `quay-monitor` label from CR overrides and test coverage for the monitoring label chain.

PR #1296, opened 2026-07-16
PROJQUAY-12322: tls: separate TLS 1.3 ciphersuites from TLS 1.2 ciphers

LiZhang19817:
## Summary

- **Bug:** When OpenShift APIServer `tlsSecurityProfile` is set to `Modern` (TLS 1.3 only), the operator passes TLS 1.3 cipher names (`TLS_AES_128_GCM_SHA256`, etc.) to Quay's `SSL_CIPHERS` config. This maps to nginx's `ssl_ciphers` directive, which calls `SSL_CTX_set_cipher_list()` — but TLS 1.3 ciphers require `SSL_CTX_set_ciphersuites()`. Nginx crashes with `no cipher match`.
- **Fix:** Split `translateTLSProfile()` to return separate TLS 1.2 ciphers and TLS 1.3 ciphersuites (distinguished by the `TLS_` prefix). TLS 1.3 ciphers are now emitted as `SSL_CIPHERSUITES` in Quay config, which maps to nginx's `ssl_conf_command Ciphersuites` directive.
- Old/Intermediate/Custom profiles continue to work correctly — TLS 1.3 ciphers in mixed lists are automatically routed to the correct directive.

**JIRA:** https://issues.redhat.com/browse/PROJQUAY-12322

## Changed Files

| File | Change |
|------|--------|
| `pkg/context/context.go` | Added `SSLCiphersuites` field to `QuayRegistryContext` |
| `controllers/quay/tls.go` | `translateTLSProfile` returns 3 values; added `splitCiphers()` / `joinCiphers()` |
| `pkg/kustomize/kustomize.go` | Inject `SSL_CIPHERSUITES` into generated Quay config |
| `controllers/quay/tls_test.go` | Updated tests for new signature; added Modern and mixed-cipher regression tests |

## Test plan

- [x] `make fmt` passes
- [x] `make vet` passes
- [x] `make test` — all unit tests pass
- [ ] Deploy on OpenShift 4.22 with APIServer `tlsSecurityProfile: Modern` — verify Quay pods start without nginx crash
- [ ] Verify Intermediate profile still works (TLS 1.2 ciphers in `SSL_CIPHERS`, TLS 1.3 in `SSL_CIPHERSUITES`)
- [ ] Verify Custom profile with mixed ciphers splits correctly
- [ ] Verify user-provided `SSL_CIPHERS`/`SSL_CIPHERSUITES` in configBundleSecret still takes precedence

🤖 Generated with [Claude Code](https://claude.ai/code)

[...]

PR #1316, opened 2026-08-25
fix(#1298): nest tools block under reviews in CodeRabbit config

### LINKED ISSUE #1298: fix .coderabbit.yaml: `tools` key is unrecognized, golangci-lint and gitleaks are silently disabled
## What happened

During the CodeRabbit review of PR #1296, CodeRabbit posted a configuration warning:

> Validation error: Unrecognized key: "tools"

The `.coderabbit.yaml` file (lines 298-304) defines a `tools:` section enabling golangci-lint and gitleaks. However, CodeRabbit does not recognize this key, meaning both integrations are silently disabled. No static analysis from golangci-lint runs during reviews, and no secret scanning from gitleaks is performed.

This is a pre-existing configuration issue, not introduced by PR #1296.

Ref: https://github.com/quay/quay-operator/pull/1296

## What could go better

golangci-lint integration in CodeRabbit reviews could catch issues like unused variables, inefficient code patterns, and potential bugs that manual review might miss. gitleaks could catch accidental secret exposure in diffs. Both are valuable review augmentations that the repo maintainers intended to enable but aren't actually running.

Confidence: High — the warning is directly visible in the PR comment from CodeRabbit, and the `tools:` key at line 298 of `.coderabbit.yaml` is clearly present but not being processed.

## Proposed change

Check the current CodeRabbit schema (https://coderabbit.ai/integrations/schema.v2.json) for the correct location of tool configuration. The `tools` section may need to be nested under `reviews:` or may have a different key name in the v2 schema. Update `.coderabbit.yaml` accordingly so that golangci-lint and gitleaks are properly recognized and enabled.

If the v2 schema no longer supports these integrations as inline config, remove the dead `tools:` section to avoid confusion and configure the tools through CodeRabbit's dashboard or alternative mechanism.

## Validation criteria

1. CodeRabbit no longer reports a configuration warning about unrecognized keys on new PRs.
2. CodeRabbit reviews include golangci-lint findings (test by submitting a PR with a known lint issue like an unused variable).
3. The `.coderabbit.yaml` file passes schema validation against https://coderabbit.ai/integrations/schema.v2.json.

---
_Generated by retro agent from https://github.com/quay/quay-operator/pull/1296_

Ryntak94:
The `tools:` block in `.coderabbit.yaml` sat at the YAML document root, as a sibling of `reviews:`, `knowledge_base:` and `chat:`. The CodeRabbit v2 schema (https://coderabbit.ai/integrations/schema.v2.json) sets `additionalProperties: false` at the root and only allows `language`, `tone_instructions`, `early_access`, `enable_free_tier`, `inheritance`, `issue_enrichment`, `code_generation`, `reviews`, `knowledge_base` and `chat` there. `tools` is defined under `reviews`, so the root-level block was rejected as an unrecognized key — which is exactly the `Validation error: Unrecognized key: "tools"` warning CodeRabbit posted on #1296 — and both the golangci-lint and gitleaks integrations were silently disabled for every review.

This moves the block to `reviews.tools`, keeping the `golangci-lint` and `gitleaks` keys (both valid per the schema, each taking an `enabled` boolean) and their explanatory comments, and relocates the section banner as a nested subsection alongside the other `reviews.*` subsections. No settings change, only their position in the document.

---

Closes #1298

### Validation

- [x] `tools:` now sits inside the `reviews:` block (`.coderabbit.yaml` line 233), where the v2 schema defines it, with the `golangci-lint` and `gitleaks` keys and their `enabled` booleans unchanged
- [x] The file validated cleanly against schema.v2.json when the change was made
- [ ] **Not verifiable on this PR** — criteria 1 and 2 in the issue (no configuration warning; golangci-lint/gitleaks findings appearing in reviews) can only be confirmed *after* this merges. CodeRabbit reviews a fork PR using the target branch's configuration, so the review here reads `master`'s root-level `tools:` key and warns about it regardless of what this branch does. See [the comment below](https://github.com/quay/quay-operator/pull/1316#issuecomment-5404607148) for the details; both criteria should be checked on the first PR opened after this one lands.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -228,6 +228,17 @@ reviews:
           environment variables, not mutable tags like :latest.
 
   # ---------------------------------------------------------------------------
+  # Tools — static analysis integrations
+  # ---------------------------------------------------------------------------
+  tools:
+    # golangci-lint — no project-level config file; uses pre-commit default settings.
+    golangci-lint:
+      enabled: true
+    # gitleaks — scan for leaked secrets/credentials in diffs.
+    gitleaks:
+      enabled: true
+
+  # ---------------------------------------------------------------------------
   # Pre-merge checks — automated validation rules
   # ---------------------------------------------------------------------------
   pre_merge_checks:
@@ -291,17 +302,6 @@ knowledge_base:
       - "GEMINI.md"
   learnings:
     scope: local
-
-# -----------------------------------------------------------------------------
-# Tools — static analysis integrations
-# -----------------------------------------------------------------------------
-tools:
-  # golangci-lint — no project-level config file; uses pre-commit default settings.
-  golangci-lint:
-    enabled: true
-  # gitleaks — scan for leaked secrets/credentials in diffs.
-  gitleaks:
-    enabled: true
 
 # -----------------------------------------------------------------------------
 # Chat — auto-reply to @coderabbitai mentions


Ryntak94:
### Note for reviewers: the `Unrecognized key: "tools"` warning above is expected on this PR

CodeRabbit has posted its usual `Validation error: Unrecognized key: "tools"` note on this PR — the very warning this PR fixes. That is not evidence the fix doesn't work. CodeRabbit reviews a fork PR using the configuration from the **target** branch, not the one in the PR. It spells this out itself on the sibling PR #1317:

> This pull request changes a CodeRabbit configuration file. Because it comes from a fork or its author is not a repository collaborator, reviews use only the configuration from the target branch. The proposed configuration will take effect after it is merged.

So the review that runs here reads `master`'s `.coderabbit.yaml`, where `tools:` is still at the document root — and correctly warns about it. The moved block on this branch is never loaded.

That makes validation criterion 1 in #1298 ("CodeRabbit no longer reports a configuration warning on new PRs") unverifiable before merge, by construction; the same applies to criterion 2, since golangci-lint and gitleaks stay disabled while the root-level `tools:` key is the one in effect. Both should be confirmed on the first PR opened **after** this one merges, and this change is trivially revertable if they aren't.

What can be checked here and now, from the diff alone:

- `tools:` now sits under `reviews:`, which is where the [v2 schema](https://coderabbit.ai/integrations/schema.v2.json) defines it — the root object sets `additionalProperties: false` and does not list `tools` among its allowed keys.
- The `golangci-lint` and `gitleaks` entries and their `enabled` booleans are carried over unchanged; this is a move, not a settings change.

PR #1317, opened 2026-08-25
NO-ISSUE: docs(#1299): distinguish Ginkgo and testing.T guidance for test files

### LINKED ISSUE #1299: refine .coderabbit.yaml test path instruction to distinguish Ginkgo vs standard testing usage
## What happened

In PR #1296, CodeRabbit posted a nitpick on `controllers/quay/tls_test.go` requesting conversion to Ginkgo v2/Gomega matchers, citing the path instruction: "controller tests use Ginkgo v2 with Gomega matchers."

However, the existing pattern in `controllers/quay/` is mixed:
- `quayregistry_controller_test.go` — uses Ginkgo v2 via the test suite (`suite_test.go`)
- `features_test.go` — uses standard `testing.T` with table-driven tests
- `tls_test.go` — uses standard `testing.T` with table-driven tests

The Ginkgo suite is used for integration-style controller tests that require envtest, while utility function tests (hostname parsing, TLS profile translation, cipher splitting) use standard `testing.T`. This is a reasonable separation, but the path instruction doesn't reflect it.

Ref: https://github.com/quay/quay-operator/pull/1296

## What could go better

The overly broad path instruction caused a false positive nitpick that, while not blocking, adds noise to reviews. If a future code agent treats this as an actionable requirement, it could waste tokens converting tests that are already appropriately written with standard `testing.T`.

Confidence: Medium — this is a minor review noise issue. The nitpick was categorized as "Trivial" with "Poor tradeoff" and didn't cause rework. But fixing the instruction would prevent recurring false positives on future PRs touching controller utility tests.

## Proposed change

Update the test file path instruction in `.coderabbit.yaml` (line 177-194) to distinguish between the two patterns:

```yaml
- path: "**/*_test.go"
  instructions: >
    Unit and integration tests.

    Review checklist:
    - Test framework: controller integration tests (those testing the
      full reconcile loop with envtest) use Ginkgo v2 with Gomega matchers.
      Utility function tests in controllers/ and other packages use
      standard testing.T with table-driven patterns — this is acceptable
      and preferred for pure functions.
    ...
```

This clarifies that standard `testing.T` is the right choice for utility tests, reducing false positive style nitpicks.

## Validation criteria

1. Future PRs modifying `controllers/quay/tls_test.go` or `controllers/quay/features_test.go` do not receive Ginkgo conversion nitpicks from CodeRabbit.
2. PRs adding new Ginkgo-based controller integration tests still pass the path instruction check.
3. The instruction change is consistent with the actual test patterns in the codebase.

---
_Generated by retro agent from https://github.com/quay/quay-operator/pull/1296_

Ryntak94:
The `**/*_test.go` path instruction said the project "uses Ginkgo v2 with Gomega matchers for controller tests, and standard testing package elsewhere". Read literally, that makes every file under `controllers/` Ginkgo territory, which is what produced the false-positive nitpick on #1296 asking for `controllers/quay/tls_test.go` to be converted to Ginkgo even though it is a pure-function table test.

What is actually in the tree: of 23 test files, only two import `onsi/ginkgo`. `controllers/quay/suite_test.go` bootstraps envtest and runs the specs, and `controllers/quay/quayregistry_controller_test.go` holds a single `Describe("Reconciling a QuayRegistry")` block. The remaining 87 top-level test functions — including 11 in that same controller test file, plus `controllers/quay/tls_test.go` and `features_test.go` — are standard `testing.T` table-driven tests using the fake client.

This rewords the instruction so the framework is tied to what a test exercises rather than to the directory it lives in: Ginkgo v2 + Gomega + envtest for controller integration tests that drive the reconcile loop, standard `testing.T` for utility and fake-client tests wherever they live, including under `controllers/quay/`. The envtest, `time.Sleep` and assertion bullets are scoped the same way so they no longer imply Gomega applies to every test file.

Config-only change; no Go code is touched.

---

Closes #1299

### Validation

- [x] Counts above taken from the tree at `master`: of the 23 `*_test.go` files, `git grep -l onsi/ginkgo -- '*_test.go'` returns exactly `controllers/quay/suite_test.go` and `controllers/quay/quayregistry_controller_test.go`
- [x] This change introduces no new schema errors: validated against [schema.v2.json](https://coderabbit.ai/integrations/schema.v2.json), both `master` and this branch produce the same single pre-existing error — `Additional properties are not allowed ('tools' was unexpected)`, from the root-level `tools:` key. That key is unrelated to this change and is fixed separately in #1316, after which the file validates with zero errors. The `path_instructions` entry edited here is schema-valid.
- [ ] The real check is behavioural — future PRs touching `controllers/quay/tls_test.go` should stop drawing Ginkgo-conversion nitpicks

## Summary by CodeRabbit

* **Documentation**
  * Clarified test review guidance for integration tests using envtest versus standard unit and utility tests.
  * Added guidance on test suite locations, cleanup requirements, and assertion practices.
  * Specified that envtest should be limited to the controller test suite.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -199,12 +199,17 @@ reviews:
           up an API server.
         - Cleanup: resources created in tests must be cleaned up via defer,
           t.Cleanup, or AfterEach. Flag leaked resources.
-        - No time.Sleep: for async assertions in the Ginkgo suite use
-          Eventually/Consistently with an explicit timeout and polling
-          interval. Flag time.Sleep anywhere.
-        - Assertion quality: Gomega assertions should include meaningful
-          context via .WithOffset() or descriptive wrappers; testing.T
-          failures should name the case and report got vs. want.
+        - No time.Sleep for asynchronous assertions: in the Ginkgo controller
+          suite use Eventually/Consistently with an explicit timeout and
+          polling interval. Flag time.Sleep that waits for asynchronous
+          behaviour; do not raise it against standard testing.T utility or
+          fake-client tests, which have nothing to wait on.
+        - Assertion quality: Gomega assertions should carry meaningful context
+          through descriptive failure messages or wrappers. .WithOffset() (and
+          the *WithOffset variants) does not add context — it only attributes
+          a failure to the caller of a shared helper, so only ask for it on
+          that basis. testing.T failures should name the case and report got
+          vs. want.
         - Table-driven tests: for functions with many input/output combinations,
           prefer table-driven patterns ([]struct{} loops in standard tests,
           DescribeTable in Ginkgo).
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -178,20 +178,36 @@ reviews:
         Unit and integration tests.
 
         Review checklist:
-        - Test framework: this project uses Ginkgo v2 with Gomega matchers
-          for controller tests, and standard testing package elsewhere. Verify
-          correct framework usage.
-        - envtest: controller tests use envtest (sigs.k8s.io/controller-runtime/pkg/envtest).
-          Verify proper test environment setup and teardown.
-        - Cleanup: resources created in tests must be cleaned up via defer or
-          AfterEach. Flag leaked resources.
-        - No time.Sleep: use gomega.Eventually/Consistently with timeout and
-          polling interval instead of time.Sleep for async assertions.
-        - Assertion messages: Gomega assertions should include meaningful
-          context via .WithOffset() or descriptive wrappers.
+        - Test framework: the framework follows what a test exercises, not the
+          directory it lives in. Controller INTEGRATION tests that drive the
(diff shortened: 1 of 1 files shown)


Ryntak94:
Thanks — both were fair, and both are fixed in 8813e1cb.

**`time.Sleep`:** you're right that `Flag time.Sleep anywhere` reached straight past the scoping this PR adds and would have kept producing the false positives it is meant to remove. The rule is now scoped to what it is actually about — sleeping instead of `Eventually`/`Consistently` while waiting for asynchronous behaviour in the Ginkgo controller suite — and explicitly does not apply to `testing.T` utility or fake-client tests, which have nothing to wait on.

**`.WithOffset()`:** correct, and the original bullet was simply wrong about what it does. It shifts which call site is reported when an assertion inside a shared helper fails; it adds no context. The bullet now asks for context via descriptive failure messages or wrappers, and describes `.WithOffset()` as the helper-attribution mechanism it is.

PR #1324, opened 2026-09-01
PROJQUAY-5850: operator: support STS configuration flow via OLM and CCO

sudipshil9862:
## Summary

- Integrate the Quay operator with OpenShift's Cloud Credential Operator (CCO) so that Quay pods on STS-enabled clusters authenticate to AWS S3 using short-lived tokens instead of static IAM access keys
- When ROLEARN env var is set via OLM subscription config, the operator creates a CredentialsRequest, CCO provisions a Secret, and the operator mounts it into quay-app/quay-mirror pods with AWS_SHARED_CREDENTIALS_FILE
- Blocks rollout when static AWS credentials conflict with STS, degrades when CCO fails to provision credentials within 5 minutes, and falls back to normal operation when ROLEARN is unset

## Key design decisions

- **Local CredentialsRequest types** (`pkg/credentialsrequest/`) instead of importing the CCO library to avoid transitive dependency conflicts (CCO@latest requires k8s v0.36, we are on v0.28)
- **CRD detection via RESTMapper at startup** (cached bool) instead of `r.List()` every reconcile — matches the existing Route API detection pattern
- **Block rollout on credential conflict** — when ROLEARN is set but static s3_access_key/s3_secret_key exist in configBundleSecret, the operator blocks with a clear message telling the admin to remove static credentials

## Files changed

| File | Change |
|------|--------|
| `pkg/credentialsrequest/types.go` | New — local CredentialsRequest types and builder |
| `pkg/context/context.go` | Add STS context fields |
| `apis/quay/v1/quayregistry_types.go` | Add condition reason constants |
| `controllers/quay/features.go` | `checkSTSCapability()` + `hasStaticAWSStorageKeys()` |
| `controllers/quay/quayregistry_controller.go` | `ensureCredentialsRequest()` + reconcile loop wiring + RBAC |
| `pkg/middleware/middleware.go` | `applySTSCredentials()` — volume/mount/env injection |
| `bundle/manifests/quay-operator.clusterserviceversion.yaml` | `token-auth-aws: "true"` + CredentialsRequest RBAC |

## Test plan

- [x] `TestCheckSTSCapability` — 5 cases: ROLEARN unset, managed objectstorage, static key conflict, CRD unavailable, happy path
- [x] `TestHasStaticAWSStorageKeys` — 7 cases: empty config, S3Storage with/without keys, RadosGWStorage, STSS3Storage, empty keys
- [x] `TestApplySTSCredentials` — 5 cases: volumes, mounts, env var, idempotency
- [x] `TestNewCredentialsRequest` — 4 cases: GVK, metadata, spec fields, providerSpec with role ARN
- [x] Full build passes (`go build ./...`)
- [x] `go vet` clean
- [x] `gofmt` clean
- [ ] E2E test on STS-enabled cluster (requires ROSA/OSD environment)

## Related

- JIRA: [PROJQUAY-5850](https://redhat.atlassian.net/browse/PROJQUAY-5850)
- Reference: Short-Lived Token Auth Guide ([OCPSTRAT-171](https://redhat.atlassian.net/browse/OCPSTRAT-171))

## Summary by CodeRabbit

* **New Features**
  * Added AWS STS authentication for Quay deployments using unmanaged S3 storage.
  * Automatically provisions short-lived credentials and applies them to Quay application and mirror components.
  * Reports credential readiness, pending requests, unavailable provisioning support, and conflicting credentials.
  * Updates credential configuration when storage or credential definitions change.

* **Documentation**
  * Added setup and configuration guidance for AWS STS, IAM permissions, and troubleshooting.

* **Tests**
  * Expanded coverage for STS detection, provisioning, validation, mounting, cleanup, and conflict handling.

### DIFF
--- a/bundle/manifests/quay-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/quay-operator.clusterserviceversion.yaml
@@ -179,12 +179,16 @@ spec:
                 - infrastructures
               verbs:
                 - get
+                - list
+                - watch
             - apiGroups:
                 - operator.openshift.io
               resources:
                 - cloudcredentials
               verbs:
                 - get
+                - list
+                - watch
             - apiGroups:
                 - networking.k8s.io
               resources:
--- a/bundle/manifests/quay-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/quay-operator.clusterserviceversion.yaml
@@ -42,7 +42,7 @@ metadata:
     features.operators.openshift.io/fips-compliant: "true"
     features.operators.openshift.io/proxy-aware: "true"
     features.operators.openshift.io/tls-profiles: "false"
-    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
   name: quay-operator.v3.99.0-dev
@@ -186,6 +186,18 @@ spec:
                 - delete
                 - update
                 - patch
+            - apiGroups:
+                - cloudcredential.openshift.io
+              resources:
+                - credentialsrequests
+              verbs:
+                - create
+                - delete
+                - get
+                - list
+                - patch
+                - update
+                - watch
           serviceAccountName: quay-operator
       permissions:
         - rules:
--- a/bundle/manifests/quay-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/quay-operator.clusterserviceversion.yaml
@@ -175,6 +175,14 @@ spec:
                 - config.openshift.io
               resources:
                 - apiservers
+                - authentications
+                - infrastructures
+              verbs:
+                - get
+            - apiGroups:
+                - operator.openshift.io
+              resources:
+                - cloudcredentials
               verbs:
                 - get
             - apiGroups:
--- a/bundle/manifests/quay-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/quay-operator.clusterserviceversion.yaml
@@ -8,7 +8,7 @@ metadata:
     containerImage: quay.io/projectquay/quay-operator:v3.99.0-dev
     createdAt: 2021-04-23T10:04:00Z
     support: Project Quay
-    description: Opinionated deployment of Quay on Kubernetes.
+    description: Opinionated deployment of Quay on Kubernetes, including AWS STS authentication for unmanaged S3 storage. See https://github.com/quay/quay-operator/blob/master/docs/sts-iam-setup.md.
     quay-version: v3.99.0-dev
     repository: https://github.com/quay/quay-operator
     tectonic-visibility: ocs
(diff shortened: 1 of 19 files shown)


sudipshil9862:
**Finding 3 — Reconcile existing STS volumes (pkg/middleware/middleware.go:716-717):**

Fixed in f65153c2. Renamed `appendVolumeIfAbsent`/`appendVolumeMountIfAbsent` to `upsertVolume`/`upsertVolumeMount` — same-name entries are now replaced with the desired definition instead of retained unchanged. Added a regression test (`replaces_conflicting_volume_definition`) that verifies a stale Secret reference and mount path are overwritten on the next reconcile.

**Finding 1 — Use versioned CCO API types (pkg/credentialsrequest/types.go:16-39):**

Intentional design decision. Importing `github.com/openshift/cloud-credential-operator` pulls in transitive dependencies that upgrade k8s.io/api from v0.28 to v0.36 and controller-runtime from v0.16 to v0.24, breaking the build. The STS guide explicitly shows a local struct approach as a valid alternative ("If you're starting from a CredentialsRequest in YAML form, you may want to handle it something like this"). The CredentialsRequest object has ~10 fields — maintaining a local struct is lower risk than coupling to CCO's release cycle.

**Finding 2 — Restrict IAM policy to configured S3 bucket (pkg/credentialsrequest/types.go:60-61):**

The `statementEntries` in the CredentialsRequest are metadata for `ccoctl` to document what permissions Quay needs — they do not enforce IAM policy. The actual permissions are on the admin-created IAM role, which the admin scopes to their bucket. The STS guide itself uses `s3:*` with `arn:aws:s3:*:*:*` in its reference implementation. The bucket name is also not available at CredentialsRequest creation time in a clean way — it lives in the user's `DISTRIBUTED_STORAGE_CONFIG` which is per-QuayRegistry configuration.

bundle/manifests/quay-operator.clusterserviceversion.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     containerImage: quay.io/projectquay/quay-operator:v3.99.0-dev
     createdAt: 2021-04-23T10:04:00Z
     support: Project Quay
-    description: Opinionated deployment of Quay on Kubernetes.
+    description: Opinionated deployment of Quay on Kubernetes, including AWS STS authentication for unmanaged S3 storage. See https://github.com/quay/quay-operator/blob/master/docs/sts-iam-setup.md.
```
Marcusk19, line 11: nit: I'm not sure where this description will show up, but do we need to call out the AWS STS feature specifically here?

pkg/credentialsrequest/types.go
```diff
@@ -0,0 +1,101 @@
...
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+type CredentialsRequestSpec struct {
+	SecretRef           SecretRef       `json:"secretRef"`
+	ProviderSpec        json.RawMessage `json:"providerSpec"`
+	ServiceAccountNames []string        `json:"serviceAccountNames"`
+	CloudTokenPath      string          `json:"cloudTokenPath"`
+}
```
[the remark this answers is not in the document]
Marcusk19: seems like this should be addressed for maintainability

Marcusk19 (commented):
overall lgtm, just one nit as an inline comment and I think the coderabbit finding should be addressed. has this been tested on a live AWS cluster yet?

bundle/manifests/quay-operator.clusterserviceversion.yaml
bcaton85: nope, good call. @sudipshil9862 can you update this?

bcaton85:
yes we deployed and tested to our development cluster. Also we're setting up e2e tests that will also test on a live cluster

